PR #83, opened 2023-08-31
[Feature] Add starcoder example

HIT-cwh:


xtuner/utils/templates.py
```diff
@@ -84,4 +84,6 @@
                            'helpful, detailed, and polite answers to the '
                            'user\'s questions. USER: {input} ASSISTANT: '),
         INSTRUCTION=('USER: {input} ASSISTANT: ')),
-)
+    stack_exchange=dict(
+        INSTRUCTION_START='Question: {input}\n\nAnswer: ',
+        INSTRUCTION='Question: {input}\n\nAnswer: '))
```
LZHgrla: it is better to add a `,` after the `)`, to facilitate future expansion.

LZHgrla (approved):

PR #244, opened 2023-11-24
[Feature] Filter negative labels

xiaohangguo:
--------------

在数据集处理时，如果训练集中`input`过长，超过了`max_length`，这种数据虽然在`tokenizer`部分被标记成了`-100`，但仍然会参与向前计算和损失函数的计算，这会浪费计算资源，因此写了一个`filter_out_all_negative_labels`函数来过滤掉负标签的样本。这样使得负样本不会参与向前传播的计算和损失函数的计算。

---------------
During dataset processing, if the `input` in the training set is too long and exceeds `max_length`, such data, although marked as `-100` in the tokenizer section, will still participate in the forward computation and loss function calculation, which can waste computational resources. Therefore, a function named `filter_out_all_negative_labels` was written to filter out samples with negative labels. This ensures that samples with negative labels do not participate in forward propagation and loss function calculations.


LZHgrla (approved):
Thanks!
LGTM

PR #266, opened 2023-12-12
[Feature]Support CEPH

pppppM:


xtuner/utils/fileio.py
```diff
@@ -0,0 +1,179 @@
...
+        if global_vars is not None and 'open' in global_vars:
+            global_vars['open'] = bak_open
+        setattr(patch_fileio, '_patched', False)
+
+
+def patch_hf_auto_from_pretrained():
+    if hasattr(patch_hf_auto_from_pretrained, '_patched'):
+        return
+
+    from transformers import AutoModelForCausalLM, AutoTokenizer
```
LZHgrla: Suggest modifying the `from_pretrained` method of these classes.

```python
from transformers import AutoConfig, AutoTokenizer, AutoImageProcessor, AutoFeatureExtractor, AutoProcessor, PreTrainedTokenizerBase, PreTrainedModel, ImageProcessingMixin, ProcessorMixin
from transformers.models.auto.auto_factory import _BaseAutoModelClass
from peft import PeftModel
```

This is an inflexible approach, but I can't think of a better solution.

HIT-cwh:
We can set HF_DATASETS_OFFLINE=True to prevent network operations during the execution of load_dataset.

xtuner/utils/fileio.py
pppppM: Done

pppppM:
Three new environment variables have been added:

- **HF_PETREL_ON**: Determines whether to automatically rewrite HF's `from_pretrained` and `save_pretrained` interfaces when importing xtuner.
- **HF_PETREL_HUB**: If not empty, it will automatically set **HF_PETREL_ON** as TRUE. Suppose HF_PETREL_HUB is 's3://xtuner' and the model card is 'internlm/internlm-7b', the final loading path would be 's3://xtuner/internlm/internlm-7b'.
- **DEEPSPEED_PETREL_ON**: Determines whether to read and store checkpoints from Peterl.

HIT-cwh (approved):

PR #339, opened 2024-01-19
Update internlm2_chat template

RangiLyu:
Replace unused token ids with readable chatml special tokens.


LZHgrla (approved):
LGTM, Thanks!

PR #385, opened 2024-02-01
[Feature] support saving eval output before save checkpoint

HIT-cwh:


xtuner/engine/hooks/evaluate_chat_hook.py
```diff
@@ -160,9 +174,13 @@ def _generate_samples(self, runner, max_new_tokens=None):
                     max_new_tokens=max_new_tokens,
                     generation_config=self.gen_config,
                     stopping_criteria=self.stop_criteria)
-                runner.logger.info(
-                    f'Sample output:\n'
-                    f'{self.tokenizer.decode(generation_output[0])}\n')
+                generation_output = self.tokenizer.decode(generation_output[0])
+                runner.logger.info(f'Sample output:\n{generation_output}\n')
+                if save_eval_output:
+                    eval_outputs.append(f'{inputs + generation_output}\n')
```
LZHgrla: Do we need to save `inputs` here？

xtuner/engine/hooks/evaluate_chat_hook.py
```diff
@@ -180,10 +198,18 @@ def after_train_iter(self,
                          data_batch=None,
                          outputs=None) -> None:
         if (self.every_n_iters is None or batch_idx == 0
-                or batch_idx % self.every_n_iters != 0):
+                or (batch_idx + 1) % self.every_n_iters != 0):
```
LZHgrla: `batch_idx + 1` will cause the execution iters of evaluating generation and saving ckpt to be inconsistent.

xtuner/engine/hooks/evaluate_chat_hook.py
HIT-cwh: Save the evaluation results for gui display

xtuner/engine/hooks/evaluate_chat_hook.py
HIT-cwh: batch_idx = runner.iter = (real_iter - 1).  (https://github.com/open-mmlab/mmengine/blob/main/mmengine/runner/loops.py#L314)

Suppose (save_steps = evaluation_freq = 5), we need to save eval results and ckpt when runner.iter equals to 4.

LZHgrla (approved):
Thanks! LGTM!

Issue #357, opened 2024-01-25
The space between words disapear


yiyexy:
<img width="1195" alt="image" src="https://github.com/InternLM/xtuner/assets/35927125/a0b539be-2cf0-4d37-ba24-7035da7e237b">

`xtuner chat internlm/internlm2-chat-7b \
  --visual-encoder openai/clip-vit-large-patch14-336 \
  --llava xtuner/llava-internlm2-7b \
  --prompt-template internlm2_chat \
  --image $IMAGE_PATH`

LZHgrla:
Hi @yiyexy  
Could you please paste your env info here, and I will try to fix it.
```
pip list
```

yiyexy:
`Package                       Version
----------------------------- ------------
accelerate                    0.26.1
addict                        2.4.0
aiohttp                       3.9.1
aiosignal                     1.3.1
annotated-types               0.6.0
async-timeout                 4.0.3
attrs                         23.2.0
bitsandbytes                  0.42.0
certifi                       2023.11.17
charset-normalizer            3.3.2
contourpy                     1.2.0
cycler                        0.12.1
datasets                      2.16.1
deepspeed                     0.12.6
dill                          0.3.7
distro                        1.9.0
einops                        0.7.0
et-xmlfile                    1.1.0
filelock                      3.13.1
fonttools                     4.47.2
frozenlist                    1.4.1
fsspec                        2023.10.0
func-timeout                  4.3.5
hjson                         3.1.0
huggingface-hub               0.20.2
idna                          3.6
importlib-metadata            7.0.1
Jinja2                        3.1.3
jsonschema                    4.21.0
jsonschema-specifications     2023.12.1
kiwisolver                    1.4.5
lagent                        0.1.2
markdown-it-py                3.0.0
MarkupSafe                    2.1.3
matplotlib                    3.8.2
mdurl                         0.1.2
mmengine                      0.10.2
mpi4py-mpich                  3.1.2
mpmath                        1.3.0
multidict                     6.0.4
multiprocess                  0.70.15
MultiScaleDeformableAttention 1.0
networkx                      3.2.1
ninja                         1.11.1.1
numpy                         1.26.3
nvidia-cublas-cu12            12.1.3.1
nvidia-cuda-cupti-cu12        12.1.105
nvidia-cuda-nvrtc-cu12        12.1.105
nvidia-cuda-runtime-cu12      12.1.105
nvidia-cudnn-cu12             8.9.2.26
[...]

PR #404, opened 2024-02-05
[Fix] Fix no space in chat output using InternLM2. (#357)

KooSung:


LZHgrla:
Hi @KooSung 

Thanks for your contribution!

I modified the code, to use `TextStreamer` of `transformers` to implement our streamer.

I have tested it on internlm1/2, qwen, baichuan1/2, chatglm2/3, llama2 models, and it works well.

Could you please double-check these modifications?

pppppM (approved):

PR #411, opened 2024-02-07
[Fix] Fix dispatch to support transformers>=4.36 & Add USE_TRITON_KERNEL environment variable

HIT-cwh:
1. Fix dispatch to support transformers >= 4.36
2. Add `USE_TRITON_KERNEL` environment variable. Triton Kernels will not be used by default. One can set `USE_TRITON_KERNEL` to 1 to use Triton Kernels, such as `USE_TRITON_KERNEL=1 xtuner train ...`.


LZHgrla (approved):
LGTM!

PR #747, opened 2024-06-05
[Fix] limiting the maximum version of transformers avoids null pointer 

Kit-Hung:
After transformers version 4.41.0, the method of obtaining the sample model cache_position was changed to cause a null pointer during finetune, limiting the maximum version of transformers avoids this problem.


pppppM:
@Kit-Hung How can this issue be reproduced? Perhaps there are other ways that could be compatible with 4.41.0?

PR #764, opened 2024-06-12
[Docs] Add RLHF Content

haoxufd:


hijkzzz:
Great MR,
Would you mind citing the source of the Ray+vLLM-based RLHF architecture - OpenRLHF at README.md?
Such as in https://github.com/InternLM/xtuner?tab=readme-ov-file#%EF%B8%8F-acknowledgement
**I noticed that most RLHF-related codes/architecture of XTuner are refactored based on OpenRLHF, the Apache license 2.0 of OpenRLHF requires the original copyright statement.**
Thanks.

PR #777, opened 2024-06-17
fix a dataset_map bug of custom_dataset

YanxingLiu:
fix a dataset_map bug of custom_dataset


hhaAndroid:
@YanxingLiu Thank you very much, this issue has been resolved.

PR #889, opened 2024-08-07
Adjust the order of InternVL dataset log printing

KooSung:
Adjust the order of InternVL dataset log printing to avoid printing the wrong `len(input_ids)`.


hhaAndroid (approved):

hhaAndroid:
@KooSung Please fix the lint

PR #1063, opened 2025-09-09
[bugFix] append eos_token_ids in rollout worker manually

YanhuiDua:
Append eos_token_ids in rollout worker according to finish reason manually, and this logic will be removed when lmdeploy supports to return stop_tokens


windreamer:
It seems LMDeploy already merged relateve PR https://github.com/InternLM/lmdeploy/pull/3941 into main branch.

pppppM (approved):

PR #1093, opened 2025-09-15
[Fix] add checkpoint options to trainer cfg

nil0x9:
1. **[Fix]** originally `checkpoint_interval` was defined and implemented in trainer but was never exposed to `Trainer.from_config`, thus any attempt to set ckpt interval from cli interface would be in vain (the program always saves ckpt only by the end of training). This PR tries to fix this issue.
~~2. **[Enhance]** `sp_size` was declared in `FSDPConfig` but was defined again in trainer args. This PR removes this redundant `sp_size`.~~

### DIFF
--- a/xtuner/v1/train/trainer.py
+++ b/xtuner/v1/train/trainer.py
@@ -132,11 +132,12 @@ class TrainerConfig(BaseModel):
     global_batch_size: int | None
     work_dir: Path | str | None = None
     log_dir: Path | str | None = None
-    sp_size: int = 1
     total_step: int | None = None
     total_epoch: int | None = None
     resume: ResumeConfig | None = None
     strict_load: bool = True
+    checkpoint_interval: int | None = -1
+    checkpoint_maxkeep: int | None = -1
     hf_interval: int | None = None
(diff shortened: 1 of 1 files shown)


xtuner/v1/train/trainer.py
```diff
@@ -313,7 +313,7 @@ def __init__(
 
         self.data_mesh = self._init_data_mesh(
             fsdp_cfg.tp_size,
-            sp_size,
+            fsdp_cfg.sp_size,
```
HAOCHENYE: Actually, `sp_size` should not be defined in `fsdp_cfg`, since it is more data-related rather than model-related (this attributed should be removed). Do not use `fsdp.sp_size` and use `sp_size` passed to trainer.

nil0x9: I guess the `sp_size` in `fsdp_cfg` should be removed then?

nil0x9:
I have reverted the changes regarding sp_size, keeping only the `checkpoint_interval` part. The resulting code is tested via CLI interface and functions as expected. @HAOCHENYE

PR #1108, opened 2025-09-18
fix ratio

hhaAndroid:
the loss calculation will convert logits to float32, so for alignment, we also need to convert it to float32 here to prevent the ratio from being 1 during rl training

### DIFF
--- a/ci/scripts/test_dapo.sh
+++ b/ci/scripts/test_dapo.sh
@@ -6,6 +6,8 @@ export PYTHONPATH=/mnt/shared-storage-user/huanghaian/code/lmdeploy/:$PYTHONPATH
 export UVICORN_LOG_LEVEL="CRITICAl"
 export PYTORCH_CUDA_ALLOC_CONF='expandable_segments:True'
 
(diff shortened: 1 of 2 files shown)


xtuner/v1/module/lm_head/lm_head.py
```diff
@@ -33,7 +33,9 @@ def forward(  # type: ignore[override]
             b = self.bias
         if loss_ctx is None:
             logits = F.linear(hidden_states, w, b)
-            return None, logits
+            # Note: the loss calculation will convert logits to float32, so for alignment,
+            # we also need to convert it to float32 here to prevent the ratio from being 1 during rl training
```
RangiLyu, line 37:
```suggestion
            # we also need to convert it to float32 to prevent the ratio from not being equal to 1 during on-policy rl training
```

pppppM:
Due to the lack of use of chunk forward when computing old_logprobs, it may cause OOM (Out of Memory).
From the perspective of inference, the logits returned need to be bf16. 
It is best to convert to float32 near the computation of logprobs.

PR #1158, opened 2025-10-20
rl: add session router to rollout controller

RangiLyu:


xtuner/v1/ray/rollout/controller.py
```diff
@@ -130,6 +188,7 @@ async def rollout(
         sample_params: Optional[SampleParams] = None,
         extra_params: dict = dict(),
         format: str = "openai",
+        session_id=None,
```
hhaAndroid:
```suggestion
        session_id: int=None,
```

xtuner/v1/ray/rollout/controller.py
```diff
@@ -15,6 +19,60 @@
...
+            else:
+                break
+        for sid in to_delete:
+            self._map.pop(sid, None)
+
+    def _evict_lru_to_capacity(self):
+        while len(self._map) > self._max_sessions:
+            self._map.popitem(last=False)
+
+    async def get_worker(self, session_id: str) -> Any:
```
hhaAndroid:
```suggestion
    async def get_worker(self, session_id: int) -> Any:
```

PR #1161, opened 2025-10-21
support training lm_head with fp32

HIT-cwh:


xtuner/v1/model/moe/moe.py
```diff
@@ -653,6 +653,10 @@ def fully_shard(
         mp_policy = MixedPrecisionPolicy(
             param_dtype=self.fsdp_config.param_dtype, reduce_dtype=fsdp_config.reduce_dtype
         )
+        if self.fsdp_config.lm_head_fp32:
```
HAOCHENYE, line 656: Declaring it when using it

PR #1169, opened 2025-10-23
[Enhance] add total_consumed_tokens to logs

nil0x9:
Changes:
- add total_consumed_tokens as it is especially convenient for pre-training .
- remove unnecessary rounding for grad_norm in json/tensorboard writing (as it introduces info loss)


nil0x9:
@HAOCHENYE one additional commit removing rounding for grad_norm in trainer (2 lines of code) :)

PR #1172, opened 2025-10-24
[bugFix] fix save_hf when sglang patch torch.reductions

YanhuiDua:


xtuner/v1/model/base.py
```diff
@@ -685,7 +686,8 @@ def _save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16)
 
         if is_fused_save_rank or is_others_save_rank:
             # save_executor = ThreadPoolExecutor(max_workers=16)
-            save_executor = ProcessPoolExecutor(max_workers=16)
+            mp_context = multiprocessing.get_context("spawn")
```
HAOCHENYE: Need to describe why we should use `spawn` here

xtuner/v1/utils/update_weights_utils.py
```diff
@@ -0,0 +1,8 @@
+from torch.multiprocessing import reductions
+
+
+def monkey_unpatch_torch_reductions():
```
HAOCHENYE, line 4: Need more comments to describe why we need to do this

PR #1180, opened 2025-10-27
[Fix] Change device check from 'gpu' to 'cuda'

HAOCHENYE:


CyCle1024:
Already done in #1223 , closed.

PR #1188, opened 2025-10-28
[Fix] Fix deepseek training 2

HAOCHENYE:


xtuner/v1/module/router/noaux_router.py
```diff
@@ -71,6 +71,38 @@ def forward(self, logits) -> RouterResults:
             noise = torch.randn_like(scores) * 50
             scores_for_choice = scores + noise
 
+        if self.n_group != self.topk_group:
+            assert len(logits.shape) == 2, (
+                f"XTuner Interval bug, invalid logits shape: {logits.shape}, expected shape with "
```
pppppM, line 76: internal?

HAOCHENYE: Since `logits` is computed by the gate with output shape `(seq_len × batch_size, hidden_states)`, I raise an `Internal Error` if `logits` is not 2D, indicating the router here was called unexpectedly.

xtuner/v1/model/moe/moe.py
```diff
@@ -375,12 +376,14 @@ def _micro_batch_forward(
                         seq_ctx=seq_ctx_list,
                     )
                 hidden_states = layer_results[: len(hidden_states_list)]
-                router_logits = layer_results[len(hidden_states_list) :]
+                router_logits = layer_results[len(hidden_states_list) : len(hidden_states_list) * 2]
```
hhaAndroid, line 379: qwen3vl_text.py 也要对应修改

PR #1193, opened 2025-10-29
[Fix] fp8 error due to torch fx bug in symbolic shapes

CyCle1024:
This bug is mentioned [here](https://github.com/pytorch/pytorch/issues/160646) . And fixed by this pytorch [PR](https://github.com/pytorch/pytorch/pull/157189).
This PR ports pytorch PR#157189 and monkey patch ShapeEnv.simplify in pytorch 2.8.

### DIFF
--- a/xtuner/v1/__init__.py
+++ b/xtuner/v1/__init__.py
@@ -1,0 +1,1 @@
+from . import patch  # noqa: F401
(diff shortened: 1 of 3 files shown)


CyCle1024:
@HIT-cwh please test this PR in real fp8 jobs.

HIT-cwh:
@CyCle1024 I tested this pr using Qwen 30B on 8 gpus, it does work!

PR #1212, opened 2025-11-03
Fix ckp maxkeep

jayhenry:


xtuner/v1/train/trainer.py
```diff
@@ -798,14 +798,18 @@ def _maybe_save(self):
                     )
                 )
 
+        if self._checkpoint_maxkeep > 0 and len(current_exp.checkpoint_list) > self._checkpoint_maxkeep:
+            deleted_checkpoints = current_exp.checkpoint_list[: -self._checkpoint_maxkeep]
```
HAOCHENYE, line 802: Using `_checkpoint_maxkeep` to slice the checkpoints is risky. For example, if `_checkpoint_maxkeep` is set to 0, no checkpoints will be removed, and if `_checkpoint_maxkeep` is set to -1, first checkpoints will be removed, if `_checkpoint_maxkeep` is set to None, all checkpoints will be removed.

PR #1214, opened 2025-11-04
[Feature] support bind_numa

HIT-cwh:


xtuner/v1/train/trainer.py
```diff
@@ -881,6 +891,41 @@ def _set_deterministic(self):
...
+            local_rank = self.rank % total_GPU_per_node
+
+            # compute numa id for each locak rank
+            per_numa = total_GPU_per_node // numa_node_num
+            numa_id = local_rank // per_numa
+
+            # bind numa node
+            schedule.run_on_nodes(numa_id)
+            memory.set_membind_nodes(numa_id)
+        except Exception:
```
HAOCHENYE, line 922: Do not use bare exception here

xtuner/v1/train/trainer.py
```diff
@@ -11,6 +11,16 @@
 from shutil import rmtree
 from typing import Annotated, Literal, Sized, cast
 
+
+try:
```
HAOCHENYE: Suggest using `device=="cuda"` to determine whether bind cpu by numa, and Add `numa` and `pynvml_utils` to pyproject.toml

xtuner/v1/train/trainer.py
```diff
@@ -881,6 +891,41 @@ def _set_deterministic(self):
...
+    def _try_bind_numa(self):
+        if not HAS_NUMA:
+            logger.info("numa module not found, skip numa binding.")
+            return
+
+        try:
+            numa_node_num = numa.info.get_max_node() + 1
+            # get total gpu number of current node
+            nvsmi = nvidia_smi.getInstance()
+            total_GPU_per_node = len(nvsmi.DeviceQuery("memory.total")["gpu"])
```
HAOCHENYE: Should we use torch device count here to get GPU per node? If `CUDA_VISIBLE_DEVICES` is set, will this get an error number?

xtuner/v1/train/trainer.py
HAOCHENYE: Move this import to function, only use numa when device is cuda

xtuner/v1/train/trainer.py
```diff
@@ -894,6 +939,8 @@ def _init_dist(self, backend: str | None = None):
             init_process_group(backend=backend)
         torch.accelerator.set_device_index(int(os.environ["LOCAL_RANK"]))
 
+        self._try_bind_numa()
```
HAOCHENYE: Calling _try_bind_numa in `Trainer.__init__`, rather than in `init_dist`

xtuner/v1/train/trainer.py
HIT-cwh: 这里只是为了防止某些意外情况导致绑核不成功，但不应该影响训练，所以 try 了一下

PR #1216, opened 2025-11-04
[UnitTest] add resume twice ut

jayhenry:


tests/train/test_trainer.py
```diff
@@ -277,21 +277,24 @@ def test_resume(self):
...
             debug=False,
-            checkpoint_interval=5,
+            checkpoint_interval=2,
+            checkpoint_maxkeep=2,
         )
 
         trainer.fit()
         dist.barrier()
+        # 0. Test checkpoint_maxkeep is consistent with meta file
+        assert len(trainer.meta.latest_exp.checkpoint_list) == trainer._checkpoint_maxkeep
```
HAOCHENYE: 1. Unit tests should clearly state the expected results in plain, readable assertions (e.g., len(trainer.meta.latest_exp.checkpoint_list) == 2), rather than relying on validation of opaque internal properties.

2. Unit tests should primarily focus on testing functionality rather than implementation details. This is because internal properties may change their semantics as code evolves, while functionality typically remains stable.

PR #1228, opened 2025-11-07
Support sync weights with flattened bucket tensor for lmdeploy

RunningLeon:
lmdeploy PR: https://github.com/InternLM/lmdeploy/pull/4109

### DIFF
--- a/tests/ray/test_update_weight.py
+++ b/tests/ray/test_update_weight.py
@@ -51,9 +51,7 @@ class TestUpdateWeight(unittest.TestCase):
             gpus_per_node=8, # gpu: 8, npu: 16
             dtype="bfloat16",
             skip_load_weights=True,
-            extra_rollout_config = {
-                "lmdeploy_backend": "pytorch",
-            }
+            context_length=256,
         )
 
         # training config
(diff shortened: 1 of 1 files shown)


xtuner/v1/rl/base/worker.py
```diff
@@ -389,7 +389,7 @@ def fit(self, data_batches: list[WorkerInputItem], rollout_idx: int):
         logger_msg = f"Rollout {rollout_idx}: "
 
         if len(rollout_logprobs_list) > 0:
-            all_diffs_tensor = torch.stack([torch.tensor(d).to(DEVICE) for d in all_diffs])  # n, 4
+            all_diffs_tensor = torch.stack([torch.tensor(d).to(DEVICE) for d in all_diffs]).to(dtype=torch.float32)  # n, 4
```
CyCle1024: why explicitly to(torch.float32)?

RunningLeon: torch.mean(x) would fail if x.dtype is torch.long. In line 375 min_diff = torch.tensor(0), the dtype is torch.long.

PR #1285, opened 2025-11-21
[Fix] Fix Division-by-zero error when warmup steps set the same as total steps

nil0x9:
Currently if user set warmup steps to be the same as total steps (which is deemed valid) for `cosine` or `linear` LR scheduler, a `division-by-zero` error will be thrown. This PR fixes this case (along w/ some lint issues).

### DIFF
--- a/xtuner/v1/train/trainer.py
+++ b/xtuner/v1/train/trainer.py
@@ -21,7 +21,7 @@ from mmengine.runner import set_random_seed
 from pydantic import BaseModel, ConfigDict, field_serializer, field_validator, model_validator
 from torch.distributed import init_process_group
 from torch.distributed.device_mesh import init_device_mesh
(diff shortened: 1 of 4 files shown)


xtuner/v1/train/trainer.py
```diff
@@ -764,7 +764,7 @@ def build_engine(
             engine.model.set_hf(model_path)
         return engine
 
-    def build_lr_scheduler(self, lr_cfg: LRConfig, scheduler_step: int) -> torch.optim.lr_scheduler.LRScheduler:
+    def build_lr_scheduler(self, lr_cfg: LRConfig, scheduler_steps: int) -> LRScheduler:
```
HAOCHENYE: Keep the signatiure of `build_lr_scheduler`

PR #1301, opened 2025-11-26
[Feature] add LoadCheckpointConfig & refactor auto_resume

jayhenry:


xtuner/v1/train/trainer.py
```diff
@@ -383,9 +398,13 @@ def __init__(
             resume_cfg = ResumeConfig()
 
         self._work_dir = self._resolve_work_dir(work_dir)
-        self._meta = self._init_xtuner_meta(self.work_dir, auto_resume=resume_cfg.auto_resume)
+        logger.warning("`resume_cfg` is deprecated, please use `auto_resume` and `load_checkpoint_cfg` instead")
+        self._auto_resume = auto_resume
+        if resume_cfg.auto_resume:
+            self._auto_resume = True
+        self._meta = self._init_xtuner_meta(self.work_dir, auto_resume=self._auto_resume)
```
HAOCHENYE, line 406: Suggest adding a backward compatible method `resolve_deprecated_resume_cfg`:

```python
auto_resume, load_checkpoint_cfg = self._resolve_deprecated_resume_cfg(resume_cfg, auto_resume, load_checkpoint_cfg)
```

This line could be removed directly when we deprecate the `resume_cfg`

xtuner/v1/train/trainer.py
```diff
@@ -892,11 +914,7 @@ def _maybe_save(self, is_snapshot: bool = False) -> bool:
 
         meta_path = self.work_dir / self._META_PATH
 
-        optimizer_path = (
-            checkpoint_path / self._SAVE_OPTIMIZER_DIR
-            if self._resume_cfg.load_optimizer_states or self._resume_cfg.load_optimizer_args
-            else None
-        )
+        optimizer_path = checkpoint_path / self._SAVE_OPTIMIZER_DIR
```
HAOCHENYE, line 918: Do we need to check the `load_checkpoint_cfg` here? Otherwise optimizer will always be saved

xtuner/v1/train/trainer.py
```diff
@@ -1394,16 +1412,24 @@ def _resolve_config_conflicts(
...
-        return resume_cfg
+        if latest_checkpoint is not None and auto_resume:
+            load_checkpoint_cfg.checkpoint_path = Path(latest_checkpoint)
+            load_checkpoint_cfg.load_optimizer_states = True
+            load_checkpoint_cfg.load_optimizer_args = True
+            load_checkpoint_cfg.load_dataset = True
+            load_checkpoint_cfg.load_scheduler = True
+        return load_checkpoint_cfg
 
     def _resume(self):
```
HAOCHENYE: Suggest renaming `_resume`

HAOCHENYE: Suggest renaming `_resume` to `_load`

xtuner/v1/train/rl_trainer.py
```diff
@@ -65,6 +65,8 @@ class RLTrainerConfig(BaseModel):
     log_dir: Path | str | None = None
     total_epochs: int
     resume_config: ResumeConfig | None = None
+    auto_resume: bool = False
+    load_checkpoint_cfg: LoadCheckpointConfig | None = None
```
HAOCHENYE: Why its default value is None here?

xtuner/v1/train/rl_trainer.py
```diff
@@ -203,7 +207,9 @@ def __init__(
         work_dir: Path | str | None = None,
         log_dir: Path | str | None = None,
         total_epochs: int,
-        resume_config: ResumeConfig | None = None,
+        resume_config: ResumeConfig | None = None,  # TODO: Removed in version 1.1.0
+        auto_resume: bool = False,
+        load_checkpoint_cfg: LoadCheckpointConfig | None = None,
```
HAOCHENYE: Same as above

xtuner/v1/train/trainer.py
```diff
@@ -1394,16 +1412,24 @@ def _resolve_config_conflicts(
...
-        resume_cfg = self._resume_cfg
+        load_checkpoint_cfg: LoadCheckpointConfig = self._load_checkpoint_cfg
```
HAOCHENYE: Is this explicitly annotation is necessary?

xtuner/v1/train/trainer.py
jayhenry: Optimizer should be saved, even though load_checkpoint_cfg.load_optimizer is False. Because if auto_resume, we need optimizer checkpoint. 
If we train a small job, we can skip save checkpoint by setting checkpoint interval to None.

HAOCHENYE (approved):

PR #1306, opened 2025-11-28
[OPT] Reduce update_weight peek memory in RL

CyCle1024:
Key Change:
1. Clean extra index_select and cat logic in `xtuner/v1/model/base.py:_get_fused_hf_param` `_get_hf_params` in RL update weights.
2. Use EP shard to reduce peek memory in transferring fused big expert params in RL update weights.
3. Add Rollout param `update_weight_bucket_size_in_gb` to config pack tensor size in RL update weights.
4. `test_update_weight` default using `ep_size=4`.
5. When `skip_load_weights == True` in RolloutConfig, lmdeploy would load weights from training workers, this would save time in large scale cluster.

### DIFF
--- a/xtuner/v1/model/base.py
+++ b/xtuner/v1/model/base.py
@@ -1174,37 +1174,27 @@ class BaseModel(nn.Module):
         fsdp_unsharded_tensor_list = []
 
         # Concatenate the tensors along the FSDP shard dim
+        fuse_without_alloc = self.FSDP_SHARD_DIM == 0 and len(_fsdp_unsharded_tensor_list) == 1
         for tensors, size in zip(_fsdp_unsharded_tensor_list, origin_fsdp_size):
-            # Tn the case of one big tensor in tensor_list, the partition of tensors are contiguous.
-            # Therefore the cat and index_select operation can be omitted,
-            # and use _fuse_contiguous_chunks_without_alloc instead to reduce device peak memory.
-            # e.g. When a fused MoE weight exceeds bucket_size given, len(tensor_list) would be 1.
-            # and fused_tensor is not None reducing peak device memory.
-            fused_tensor = self._fuse_contiguous_chunks_without_alloc(tensors)
-            if fused_tensor is not None and fused_tensor.shape[self.FSDP_SHARD_DIM] == size:
-                fsdp_unsharded_tensor_list.append(fused_tensor)
-                continue
-            elif fused_tensor is not None:
-                # free memory ASAP
-                del fused_tensor
-            tensor = torch.cat(tensors, dim=self.FSDP_SHARD_DIM)
-            cat_tensor = torch.index_select(
-                tensor,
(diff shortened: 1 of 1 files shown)


xtuner/v1/model/base.py
```diff
@@ -1157,6 +1160,48 @@ def _fsdp_foreach_allgather(
 
         return fsdp_unsharded_tensor_list
 
+    @staticmethod
+    def fuse_contiguous_chunks_without_alloc(tensors: list[torch.Tensor]) -> torch.Tensor | None:
```
HAOCHENYE: It should be a private method

xtuner/v1/model/base.py
```diff
@@ -1157,6 +1160,48 @@ def _fsdp_foreach_allgather(
...
+        """Fuse contiguous chunks without extra memory allocation.
+
+        Return None if not possible.
+        """
+        if not tensors:
+            return None
```
HAOCHENYE: Why not raise an error here?

xtuner/v1/model/base.py
```diff
@@ -1141,6 +1136,14 @@ def _fsdp_foreach_allgather(
 
         # Concatenate the tensors along the FSDP shard dim
         for tensors, size in zip(_fsdp_unsharded_tensor_list, origin_fsdp_size):
+            # special case for partition of tensors are contiguous
```
HAOCHENYE: The comment should describe **why** rather than **how**

hhaAndroid (approved):

xtuner/v1/model/base.py
```diff
@@ -1157,6 +1160,48 @@ def _fsdp_foreach_allgather(
...
+
+        chunks = []
+        for t in tensors:
+            if (
+                t.untyped_storage().data_ptr() != storage.data_ptr()
+                or t.dtype != dtype
+                or t.device != device
+                or t.stride()[1:] != inner_stride
+            ):
+                return None
```
HAOCHENYE: If we made it a private function, we should remove some unnecessary check

xtuner/v1/rl/base/worker.py
```diff
@@ -597,19 +598,51 @@ def _update_weights_hf_generator(self):
...
         )
-        for name_list, param_list in chain(same_gen, fused_gen, shard_gen):
+
+        for name_list, fused_param_list in fused_gen:
+            state_dict = {name: param.detach() for name, param in zip(name_list, fused_param_list)}
+            if model.fsdp_config.ep_size > 1:
+                ep_mesh: DeviceMesh = model.ep_mesh
+                ep_size = ep_mesh.size()
+                ep_group = ep_mesh.get_group()
+                ep_rank = dist.get_rank(group=ep_group)
```
HAOCHENYE: Add comments describing what happened here.

HAOCHENYE (approved):

PR #1311, opened 2025-12-01
fix unit test: test_resume_and_load_checkpoint_cfg

jayhenry:


tests/train/test_trainer.py
```diff
@@ -468,81 +472,100 @@ def test_resume_and_load_checkpoint_cfg(tmp_path: Path):
...
-        work_dir=work_dir,
-        auto_resume=auto_resume,
-        load_checkpoint_cfg=load_checkpoint_cfg,
+        total_step=10,
+        work_dir=work_dir0,
     )
+    trainer0.fit()
+    # saved two checkpoints at step 5 and 10
+    print(f"trainer0.meta.latest_exp.checkpoint_list: {trainer0.meta.latest_exp.checkpoint_list}")
+    assert len(trainer0.meta.latest_exp.checkpoint_list) == 2
```
HAOCHENYE, line 490: Suggest using `self.assertEqual`, this can indicate the mismatch value when the test failed.

tests/train/test_trainer.py
```diff
@@ -468,81 +472,100 @@ def test_resume_and_load_checkpoint_cfg(tmp_path: Path):
...
 
-    # 3. check: auto_resume does not overwrite load_checkpoint_cfg at first time
-    assert trainer._load_checkpoint_cfg.load_dataset is False
-    assert trainer._load_checkpoint_cfg.load_scheduler is False
+    # 1. create: first train with auto_resume and load_checkpoint_cfg from trainer0's checkpoint
+    checkpoint_path = Path(trainer0.meta.latest_exp.checkpoint_list[-1])
```
HAOCHENYE:
```suggestion
    checkpoint_path = Path(trainer0.meta.latest_checkpoint)
```

tests/train/test_trainer.py
jayhenry: pytest will print mismatch value

PR #1346, opened 2025-12-09
Speed up image processing and fix training deadlock

hhaAndroid:
1 Instead of using independent preprocessing logic for single images, processing is accelerated by grouping them. 
2 Fixed the issue where the backward would block when some rank had no image data.

### DIFF
--- a/xtuner/v1/train/rl_trainer.py
+++ b/xtuner/v1/train/rl_trainer.py
@@ -235,19 +235,11 @@ class RLTrainer:
         self._rl_trainer_cfg = trainer_cfg
         self._load_from = Path(load_from) if isinstance(load_from, str) else load_from
 
-        is_hf_path = is_hf_model_path(load_from) if load_from is not None else False
-        if isinstance(is_hf_path, tuple):
-            raise NotImplementedError(is_hf_path[1])
-        else:
-            self._load_from_hf = is_hf_path
+        is_hf_path, error_info = is_hf_model_path(load_from) if load_from is not None else False, ""
+        self._load_from_hf = is_hf_path
 
         if not self._load_from_hf:
-            raise NotImplementedError
-
-        if not self._load_from_hf:
-            assert hf_interval is None and hf_max_keep is None, (
-                "`hf_interval` and `hf_max_keep` should be None when `load_from` is not a Huggingface model path, "
-            )
+            raise NotImplementedError(error_info)
 
         self._hf_max_keep = hf_max_keep
         self._hf_interval = hf_interval
--- a/xtuner/v1/train/trainer.py
+++ b/xtuner/v1/train/trainer.py
@@ -492,11 +492,8 @@ class Trainer:
         self._profile_memory = profile_memory
         self._load_from = Path(load_from) if isinstance(load_from, str) else load_from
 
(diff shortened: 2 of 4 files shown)


xtuner/v1/utils/misc.py
```diff
@@ -119,14 +119,13 @@ def record_git_info(staged_path: Path, unstaged_path: Path) -> Annotated[str, "C
     return commit
 
 
-def is_hf_model_path(path: str | Path) -> bool:
+def is_hf_model_path(path: str | Path) -> bool | tuple[bool, Exception]:
```
HAOCHENYE:
```suggestion
def is_hf_model_path(path: str | Path) -> Literal[True] | tuple[bool, Exception]:
```

xtuner/v1/train/rl_trainer.py
```diff
@@ -234,7 +234,13 @@ def __init__(
 
         self._rl_trainer_cfg = trainer_cfg
         self._load_from = Path(load_from) if isinstance(load_from, str) else load_from
-        self._load_from_hf = load_from is not None and is_hf_model_path(load_from)
+
+        is_hf_path = is_hf_model_path(load_from) if load_from is not None else False
+        if isinstance(is_hf_path, tuple):
+            raise NotImplementedError(is_hf_path[1])
+        else:
+            self._load_from_hf = is_hf_path
```
HAOCHENYE:
```suggestion
            self._load_from_hf = True
```

xtuner/v1/train/trainer.py
```diff
@@ -491,13 +491,22 @@ def __init__(
         self._profile_time = profile_time
         self._profile_memory = profile_memory
         self._load_from = Path(load_from) if isinstance(load_from, str) else load_from
-        self._load_from_hf = load_from is not None and is_hf_model_path(load_from)
+
+        is_hf_path = is_hf_model_path(load_from) if load_from is not None else False
+        if isinstance(is_hf_path, tuple):
+            self._load_from_hf = False
+        else:
+            self._load_from_hf = is_hf_path
```
HAOCHENYE:
```suggestion
        self._load_from_hf = not isinstance(is_hf_path, tuple)
```

xtuner/v1/train/trainer.py
```diff
@@ -491,13 +491,22 @@ def __init__(
...
         if not self._can_save_hf:
-            assert hf_interval is None and hf_max_keep is None, (
-                "`hf_interval` and `hf_max_keep` should be None when `load_from` is not a Huggingface model path, "
+            assert_info = (
+                f"`hf_interval`: {hf_interval} and `hf_max_keep`: {hf_max_keep} "
+                f"should be None when `load_from` is not a Huggingface model path, "
             )
+            if isinstance(is_hf_path, tuple):
+                assert_info += f", HF path load error Info: {is_hf_path[1]}"
+            assert hf_interval is None and hf_max_keep is None, assert_info
```
HAOCHENYE:
```suggestion
            if not self._load_from_hf:
                assert_info += f", HF path load error Info: {is_hf_path[1]}"
            assert hf_interval is None and hf_max_keep is None, assert_info
```

Besides, we should avoid to use the "magic" number like `1`

HAOCHENYE (approved):

PR #1374, opened 2025-12-18
add default lmdeploy/memory/ray log dir in RL scripts

YanhuiDua:


examples/v1/scripts/run_rl_submit.sh
```diff
@@ -29,7 +29,13 @@ export DATA_PATH=$DATA_PATH
 export EVAL_DATA_PATH=$EVAL_DATA_PATH
 export XTUNER_USE_FA3=${XTUNER_USE_FA3:-1}
 export XTUNER_LOG_LEVEL=${XTUNER_LOG_LEVEL:-"INFO"}
+export LMDEPLOY_LOG_FILE="${WORK_DIR}/lmdeploy_log_${current_time}.txt"
```
CyCle1024: This shell does not declare `current_time` variable at beginning.
See https://github.com/InternLM/xtuner/blob/42670fe432d942f723383869583c50e215512d29/examples/v1/scripts/run_rl_submit.sh#L97

PR #1383, opened 2025-12-22
[Fix] npu resources scripts and torch_compile_cfg

CyCle1024:
## Motivation
Some feature are not currently compatitable with npu device. This would lead to errors in RL training.

## Key Changes
1. Add `ACCELERATOR` and `ACCELERATOR_PER_NODE` parameter in `examples/v1/scripts/run_rl.sh` and `examples/v1/scripts/run_rl_submit.sh` to compatiable with npu device.
2. Make `xtuner.v1.model.base.BaseModel._resolve_compile_cfg` return empty dict when using npu device, this imples we shouldn't use torch.compile on npu device for now.


YanhuiDua (approved):
LGTM

PR #1385, opened 2025-12-23
[Enhance] add training config printing to trainer

HAOCHENYE:


xtuner/v1/train/trainer.py
```diff
@@ -689,7 +689,8 @@ def from_config(cls, config: TrainerConfig) -> Self:
             trainer_cfg=config,
             internal_metrics_cfg=config.internal_metrics_cfg,
         )
-        self.config = config
+        self._config = config
+        self._print_training_config()
```
CyCle1024, line 691: If someone is using `Trainer` without `from_config` method, the config would not be printed. Is this the desired behavior?

HAOCHENYE: What should be printed when `Trainer` is initialized without the `from_config` method? Otherwise, print the components of `Trainer` individually

CyCle1024 (approved):

PR #1425, opened 2026-01-15
[Fix] `port already in use` occured occasionally

CyCle1024:
## Motivation
In current CI rollout unittest, OSError `port already in use` occured occasionally. After several attempts, we found the main cause of this error is port tested in `find_master_addr_and_port` may not be freed after the funciton call. We can use `socket.SO_REUSEADDR` option to make the port available.

## Key Change:
1. Refactor find_master_addr_and_port with less try-catch structure
2. set sock option socket.SO_REUSEADDR to socket

### DIFF
--- a/xtuner/v1/ray/utils.py
+++ b/xtuner/v1/ray/utils.py
@@ -94,18 +94,23 @@ def find_master_addr_and_port(nums=1, start_port=None, end_port=None):
     if start_port is None:
         for _ in range(nums):
             s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            # socket.SO_REUSEADDR can help avoid TIME_WAIT state issues
+            # if the port is binded and listened by this socket and then we close it,
+            # socket.SO_REUSEADDR would make the port be reusable even it's in TIME_WAIT state.
             s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
             sockets.append(s)
             if _is_port_available(check_socket=s, port=0):
                 ports.append(s.getsockname()[1])
     else:
-        assert isinstance(start_port, int)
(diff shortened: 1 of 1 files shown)


xtuner/v1/ray/utils.py
```diff
@@ -71,48 +80,46 @@ def find_master_addr_and_port(nums=1, start_port=None, max_tries=1024):
...
-            if len(ports) < nums:
-                raise RuntimeError(
-                    f"Could not find {nums} available ports starting from port {start_port} after {max_tries} tries."
-                )
-    finally:
-        for s in sockets:
-            s.close()
+    else:
+        assert isinstance(start_port, int)
+        assert isinstance(end_port, int)
```
YanhuiDua: 有没有一种情况是：start_port is not None, end_port is None 呢？

xtuner/v1/ray/utils.py
```diff
@@ -71,48 +80,46 @@ def find_master_addr_and_port(nums=1, start_port=None, max_tries=1024):
...
-                    candidate_port += 1
-                    s.close()
-                    continue
+    ports: list[int] = []
+    sockets: list[socket.socket] = []
+
+    if start_port is None:
+        for _ in range(nums):
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            # socket.SO_REUSEADDR can help avoid TIME_WAIT state issues
```
YanhuiDua: `socket.SO_REUSEADDR `can help to bind the port in TIME_WAIT state immediately

xtuner/v1/ray/utils.py
YanhuiDua: 如果不允许start_port is not None, end_port is None的话，这里也要检查下(end_port - start_port) > nums

PR #1481, opened 2026-02-06
[Fix] make BaseModel._resolve_compile_cfg works.

CyCle1024:
## Motivation
Currently, the default value of FSDPConfig.torch_compile is `False`, and in `BaseModel._resolve_compile_cfg` would make the compile_cfg to be `{}` (empty dict) when `FSDPConfig.torch_compile==False`. 

## Key Change
We just make the default value of FSDPConfig.torch_compile into `True`.


HIT-cwh:
This PR modifies the default value of torch_compile in FSDPConfig, which will introduce a bc break.

Meanwhile, FSDPConfig.torch_compile is a deprecated key. It is suggested to set its default value to None. If users explicitly set FSDPConfig.torch_compile, a warning should be emitted, and the content of compile_cfg will be automatically updated accordingly.

PR #1486, opened 2026-02-06
feat: support fsdp2 muon optimizer

RangiLyu:
image


nil0x9:
It might be more desirable to introduce some mechanism to separate fused params IMHO? For example, MoE projections are implemented with `GroupedLinear` in XTuner, where 2D param actually means a bundle of linear weights. In this case, it might be more beneficial to split this single param tensor and treat them as different projections in NS iteration.

HAOCHENYE:
@claude

HAOCHENYE:
@claude

HAOCHENYE:
@claude Please modify this PR as you comment

PR #1521, opened 2026-03-03
Introduce RouterJudger and split judger configs into Native/Router

YanhuiDua:


xtuner/v1/ray/judger/dapo_math.py
```diff
@@ -291,8 +291,9 @@ def compute_reward(response, label, extra_info):
     return {"score": reward, "acc": out["acc"]}
 
 
-class DapoMathJudgerConfig(NativeJudgerConfig):
-    model_config = ConfigDict(extra="forbid")
+class _DapoMathJudgerDefaults:
```
jayhenry: 应该继承 BaseModel?

PR #1546, opened 2026-03-06
Fix dense rope

hhaAndroid:
## Root Cause

The `Dense` model overrides `_apply()` to unconditionally force `self.rotary_emb.to(torch.float32)` after every device/dtype conversion (e.g., `model.to(torch.bfloat16)`). This is unnecessary and harmful because:

1. `inv_freq` in `RotaryEmbedding` is registered as a **non-persistent buffer** (`persistent=False`), so it is intentionally excluded from `_apply()` transformations.
2. The `RotaryEmbedding.forward()` already handles dtype internally — it computes in `float32` via `torch.autocast(enabled=False)` and casts the output to match the input tensor's dtype.
3. Forcing the entire `rotary_emb` module to `float32` after `_apply()` conflicts with mixed-precision training and FSDP, where modules are expected to respect the requested dtype.

No other model in the codebase (e.g., Qwen3-VL, Mixtral, DeepSeek-V2) overrides `_apply()` for this purpose — they all rely on the rope module's built-in float32 computation logic.

## Fix

Remove the `_apply()` override from `Dense`, allowing PyTorch's default buffer management to work as designed. The rotary embedding's forward pass already guarantees numerical precision by computing in `float32` internally.

## Test Plan

- Verified that `RotaryEmbedding.forward()` produces correct `cos`/`sin` values after `model.to(torch.bfloat16)` without the `_apply` override.
- Confirmed model forward pass works correctly under mixed-precision training.

### DIFF
--- a/xtuner/v1/datasets/sft_tokenize_fn/openai.py
+++ b/xtuner/v1/datasets/sft_tokenize_fn/openai.py
@@ -42,7 +42,6 @@ class OpenaiTokenizeFunction(CachableTokenizeFunction[DataItem]):
             tools = item["tools"]
         if isinstance(item, dict) and "messages" in item:
             item = item["messages"]
-
         messages = ChatMessages(messages=item, tools=tools)
         tokenized = messages.tokenize(self.tokenizer, self.chat_template)
 


HAOCHENYE:
@claude review

HAOCHENYE:
@claude please help me to update the PR description

HAOCHENYE:
@claude please help me to update the PR description

HAOCHENYE:
@claude The PR description is not accuraty. But anyway, please squash merge this PR into main branch.

PR #1619, opened 2026-03-23
[Feature]: add ViT activation_offload for InternS1

NengXu001:
Reduced InternS1 training memory usage by ~10GB through ViT offloading; implemented module name configuration for ViT and MoE layers within activation offloading.

### DIFF
--- a/xtuner/v1/model/compose/intern_s1/modeling_vision.py
+++ b/xtuner/v1/model/compose/intern_s1/modeling_vision.py
@@ -36,6 +36,8 @@ from xtuner.v1.ops.others import Dropout
 from xtuner.v1.ops.act_fn import get_act_fn
 from xtuner.v1.utils import get_logger
 from xtuner.v1.module import AttnOutputs
+import os
+from xtuner.v1.utils.activation_offload import async_save_on_cpu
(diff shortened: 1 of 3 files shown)


xtuner/v1/model/compose/intern_s1/intern_s1_config.py
```diff
@@ -49,6 +49,7 @@ class InternS1VisionConfig(XTunerBaseModelConfig):
     use_mask_token: bool = False
     use_mean_pooling: bool = True
     attn_impl: Literal["flash_attention", "flex_attention", "eager_attention"] = "flash_attention"
+    text_hidden_layers: int = 0
```
HAOCHENYE: We should not modify other modules to work around functional deficiencies in `ActivationOffload` itself. I understand that, from VisionConfig's perspective, it should not need to be aware of `text_hidden_layers`.

HAOCHENYE (commented):
We should enhance the implementation of activation offload to make it more general-purpose, so that it can serve multiple models without needing to be aware of layer counts.

CyCle1024 (approved):
LGTM

PR #1630, opened 2026-03-24
[Fix] memory leak after the fisrt moe decoder layer

NengXu001:
Forces a gc to prevent extra memory occupation after the fisrt moe decoder layer.


xtuner/v1/module/decoder_layer/moe_decoder_layer.py
```diff
@@ -586,6 +588,7 @@ def _micro_batch_forward(
 
         router_logits = [router_results["logits"] for router_results in router_results_list]
         router_weights = [router_results["router_weights"] for router_results in router_results_list]
+        run_gc_once()
```
HAOCHENYE, line 591: Please test directly on the NPU with the main branch to see if the memory leak issue still exists. If it does, we may need to figure out why reference count-based reclamation is failing.

PR #1652, opened 2026-04-02
Track consumed samples in DataLoader; skip per-step total-token reduction

jayhenry:
This PR refactors how **total consumed samples** are tracked and resumed:

- Move ownership of **total consumed samples** to **`DataLoader`** (with updated save/restore paths); remove the older resume helpers from `resume.py`.
- **Fix Token accounting:** previously, **total consumed tokens** used a **global** all-reduce, which **over-counted tokens when sequence parallel (SP) is enabled**. Totals are now reduced **on the `dp_mesh`** only, matching DP-local semantics.
- **Stop reducing total tokens on every training step** to speedup e2e tgs


xtuner/v1/train/trainer.py
```diff
@@ -1209,10 +1210,11 @@ def _maybe_save(self, is_snapshot: bool = False) -> bool:
 
         return True
 
-    def _save_dataloader(self, dataloader_path: Path | str):
+    def _save_dataloader(self, dataloader_path: Path | str) -> int:
```
HAOCHENYE: `_save_dataloader` should only `save` rather than return the `total_consumed_steps`

xtuner/v1/train/trainer.py
```diff
@@ -1444,41 +1446,39 @@ def _compute_performance_metrics(
...
         e2e_train_time = self._train_time + self._train_time_offset
-        total_consumed_tokens_per_rank = self._total_consumed_tokens / self.world_size
 
         tgs = local_step_consumed_tokens / step_time
-        e2e_tgs = total_consumed_tokens_per_rank / e2e_train_time
-        exp_tgs = self._exp_consumed_tokens / self.world_size / self._train_time
+        approximate_total_consumed_tokens = (
+            self._init_total_tokens + self._local_total_consumed_tokens * self.world_size
+        )
+        approximate_total_consumed_tokens_per_rank = approximate_total_consumed_tokens / self.world_size
```
HAOCHENYE, line 1466: `approximate_total_consumed_tokens_per_rank` could be incorrect if `world_size` changed

xtuner/v1/datasets/sampler.py
```diff
@@ -137,6 +139,8 @@ def load_state_dict(self, state_dict) -> None:
         Args:
             state_dict (dict): The state of the sampler.
         """
+        tc = int(state_dict.get("total_consumed_steps", 0))
```
HAOCHENYE: Avoid single charactor short abbr

HAOCHENYE (approved):

PR #1656, opened 2026-04-06
[Fix] fix SP for InternS1 VL RL

tina-wen:
## Root Cause
Under sequence parallelism, entropy statistics and GRPO batch loss calibration were computed from sharded tensors, which could produce incorrect token counts and inconsistent metrics across SP ranks. In addition, GRPO loss batching needed to accept the forwarded SP context from the worker.

## Fix
Gather shifted labels and logprobs before entropy aggregation under SP, pass sp_mesh into RL loss batching from the training worker, and make GRPO batch construction use the SP-aware token counting path.

### DIFF
--- a/xtuner/v1/rl/base/worker.py
+++ b/xtuner/v1/rl/base/worker.py
@@ -28,6 +28,7 @@ from xtuner.v1.engine.train_engine import TrainEngine, TrainStepInfo
 from xtuner.v1.float8.float8_handler import Float8Handler
 from xtuner.v1.loss import CELossConfig, LogProbConfig
 from xtuner.v1.loss.ce_loss import CELossContext
+from xtuner.v1.loss.utils import sp_gather
 from xtuner.v1.model.base import BaseModel as XtunerBaseModel
 from xtuner.v1.model.base import ModelItem, TransformerConfig
 from xtuner.v1.model.compose.base import BaseComposeConfig, BaseComposeModel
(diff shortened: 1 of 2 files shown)


xtuner/v1/rl/base/worker.py
```diff
@@ -89,14 +90,19 @@ def calculate_entropy(
...
     global_grad_tokens: torch.Tensor,
+    sp_mesh: DeviceMesh | None = None,
 ) -> torch.Tensor | None:
     if len(old_logprobs_list) == 0 or old_logprobs_list[0] is None:
         return None
     sum_entropy: torch.Tensor | None = None
     for i, shifted_labels in enumerate(shifted_labels_list):
+        old_logprobs = old_logprobs_list[i]
+        assert old_logprobs is not None
+        if sp_mesh is not None:
```
CyCle1024, line 101: each sp rank should calculate its own rank loss, this change makes all ranks in a sp_group caculate the same loss, and it would be all_reduced sum later, thus the `sum_entropy` would times `sp_size`

xtuner/v1/rl/grpo/loss.py
```diff
@@ -75,15 +77,32 @@ def __init__(self, loss_cfg: GRPOLossConfig, loss_kwargs: GRPOLossKwargs):
...
+        rank_grad_tokens: torch.Tensor | None = None
+        for shifted_labels in shifted_labels_list:
+            if sp_mesh is not None:
+                # gather shifted_labels from different sp ranks to compute the correct loss weight
+                shifted_labels = sp_gather(shifted_labels, sp_mesh=sp_mesh, dim=1)
+            rank_grad_tokens = (
+                (shifted_labels != loss_cfg.ignore_idx).sum()
+                if rank_grad_tokens is None
+                else rank_grad_tokens + (shifted_labels != loss_cfg.ignore_idx).sum()
+            )
```
CyCle1024, line 103: All sp ranks in one sp group get the same rank_grad_tokens. Same problem above.

CyCle1024 (changes requested):

CyCle1024:
@tina-wen reference code block: https://github.com/InternLM/xtuner/blob/fad9de6ad4024c597cded8897873ecc8e9623861/xtuner/v1/loss/ce_loss.py#L141-L161 not the same compute logic, but the sp processing.

PR #1659, opened 2026-04-08
【CI】update threshold

kkscilife:
1. update threshold for case of 16nums vs 8nums
2. set compile false on qwen3.5 vl avoiding of random


kkscilife:
@claude review

PR #1673, opened 2026-04-13
[Fix] Fix hf cache monkey patch failure when HF_CACHE set to shared storage

nil0x9:
## Root Cause
The original implementation only used PID to distinguish processes when setting `HF_MODULES_CACHE`. In multi-node distributed training scenarios (e.g., with `torchrun --nnodes=N`), processes on different nodes may have the same PID. If a shared filesystem (e.g., NFS) is deployed and `HF_HOME` is set to some directory on the shared file system, chances are that the same cache directories might be used and read/write conflicts incur.

## Fix
- Changed cache directory naming from `modules_pid_{pid}` to `modules_cache_{hostname}_{pid}`

This ensures cache directory uniqueness across both single-node multi-GPU and multi-node distributed training scenarios.

### DIFF
--- a/xtuner/v1/utils/misc.py
+++ b/xtuner/v1/utils/misc.py
@@ -1,4 +1,5 @@
 import os
+import socket
 import sys
 import threading
 from functools import reduce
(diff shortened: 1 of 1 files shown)


nil0x9:
@claude review

PR #1681, opened 2026-04-16
feat: add async HuggingFace export support

caixianzhang:
# Summary

This PR adds the `async_hf_export` runtime implementation in XTuner and the corresponding unit tests.

The goal is to reduce foreground training stalls at HF export points while keeping the final exported HF directory consumable in the same shape as the existing sync HF path.

Main runtime changes are in:

- `xtuner/v1/train/trainer.py`
- `xtuner/v1/engine/train_engine.py`
- `xtuner/v1/model/base.py`

Tests are added in:

- `xtuner/tests/train/test_trainer_async_hf.py`

# What This PR Adds

## 1. Async HF runtime path

This PR adds a dedicated async HF export flow controlled by `async_hf_export`.

At a high level, the flow is:

1. `Trainer._maybe_async_save_hf()` checks whether the current step hits the HF export interval.
2. The parent process prepares a rank-local CPU snapshot for HF export.
3. A child process writes `.safetensors` files in the background.
4. The parent process continues training immediately.
5. On the next HF trigger or at the end of training, the parent waits and finalizes the pending export.

The async HF path is intentionally separate from async checkpoint:

- `async_checkpoint` controls training-state checkpointing
- `async_hf_export` controls HF export only

## 2. Prepare / writer split

This PR adds a prepare/writer split in `TrainEngine`:

- `prepare_async_hf_snapshot(hf_dir=...)`
  - iterates HF save chunks from `BaseModel._iter_hf_save_chunks(...)`
  - refreshes the long-lived `_async_hf_tensor_cache`
  - records the rank-local `weight_map`
- `write_async_hf_snapshot()`
  - runs in the forked child process
  - writes rank-local `.safetensors`
  - writes `async-hf-writer-status-rank-xxxxx-of-yyyyy.json`

This keeps the expensive file-writing portion off the foreground training path.

## 3. Commit semantics

The async HF path uses explicit commit semantics:

- export first writes into `hf-<step>.incomplete`
- parent waits for child completion and gathers all-rank writer status
- only if all ranks succeed:
  - rank0 writes HF index/config
  - tokenizer files are saved when applicable
  - `hf-<step>.incomplete` is renamed to `hf-<step>`
  - `hf-latest` is updated
  - `.xtuner` metadata is updated

This avoids exposing half-written HF exports as committed artifacts.

## 4. Reuse of latest sync HF infrastructure

The async HF path is built on the latest sync HF helpers in `BaseModel`, including:

- `_iter_hf_save_chunks(...)`
- `_write_hf_save_plan(...)`
- `_write_hf_index_and_config(...)`

As a result, async HF inherits the current sync HF save semantics, including newer behaviors around:

- `fp32_keys_pattern`
- `_get_save_dtype()`
- ignored/FSDP-excluded parameters in HF export

## 5. Retention behavior

This PR also wires async HF into the existing retention/update flow:

- `hf_max_keep` cleanup works for async HF exports
- old HF export directories are removed during finalize
- `meta.latest_exp.hf_checkpoint_list` stays consistent with on-disk state
- `hf-latest` points to the latest committed async HF export

# Tests

Added tests:

- `TestTrainerAsyncSaveHF.test_async_save_hf_interval`
- `TestTrainerAsyncSaveHF.test_async_save_hf_raises_on_writer_failure`

Covered behaviors:

- async HF export triggers at the expected save steps
- pending async HF jobs are finalized correctly
- per-rank writer status is merged into the final HF `weight_map`
- rank0 writes final HF index/config artifacts
- `hf_max_keep` cleanup behaves as expected
- `hf-latest` points to the newest committed export
- writer failure raises a global consistency error
- pending async HF state is cleared after failure

Test note:

- the fake trainer config disables sync checkpoint saving by default:
  - `checkpoint_interval=None`
  - `snapshot_interval=None`
  This keeps the unit test focused on async HF and avoids requiring a fake `save_dcp()` implementation in the mock engine.

Executed with:

```bash
cd /mnt/shared-storage-user/ailab-sys/zhangcaixian/my_xtunner/upload/xtuner
XTUNER_TEST_WORLD_SIZE=2 python3 -m pytest tests/train/test_trainer_async_hf.py -v
```

Result:

```text
2 passed
```

# Scope

This PR adds:

- async HF runtime implementation
- async HF unit tests

This PR does not change:

- sync HF export behavior
- checkpoint save/load behavior outside the trainer integration points needed for async HF
- existing `tests/train/test_trainer.py`


HAOCHENYE:
@claude review

tests/train/test_trainer_async_hf.py
```diff
@@ -0,0 +1,255 @@
...
+            )
+
+
+def prepare(fn):
+    def wrapper(self, *args, **kwargs):
+        self.alpaca_path = Path(__file__).resolve().parents[1] / "resource" / "openai_sft.jsonl"
+        self.tokenizer_path = None
+        self.temp_dir = tempfile.TemporaryDirectory()
+        self.fake_hf_model_dir = Path(self.temp_dir.name) / "fake_hf_model"
+        self.work_dir = Path(self.temp_dir.name) / "work_dir"
```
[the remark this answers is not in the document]
caixianzhang: This was true for the previous fake engine implementation. In the current version, async HF writer status file handling has moved into BaseModel, and the trainer test fake no longer defines or parses
_async_hf_writer_status_filename. The only remaining filename format is the real implementation in BaseModel, which uses the dash-separated, zero-padded format. So this nit is no longer applicable.

xtuner/v1/model/base.py
```diff
@@ -63,6 +67,55 @@
 DEVICE = get_device()
 
 
+def _set_process_qos(cpu_priority: int, io_priority: Optional[int]) -> None:
```
HAOCHENYE: move this utils function to `xtuner/v1/utils`

xtuner/v1/model/base.py
```diff
@@ -63,6 +67,55 @@
...
+                logger.debug(f"PID {pid}: CPU nice already at target {cpu_priority}")
+            elif increment < 0:
+                logger.warning(
+                    f"PID {pid}: Skipping CPU nice change from {current_nice} to {cpu_priority}; "
+                    "lowering nice requires privilege"
+                )
+            else:
+                new_nice = os.nice(increment)
+                logger.debug(f"PID {pid}: Set CPU nice from {current_nice} to {new_nice} (target {cpu_priority})")
+        except (OSError, PermissionError) as e:
```
HAOCHENYE: PermissionError is s subclass of OSError

```suggestion
        except OSError as e:
```

xtuner/v1/model/base.py
```diff
@@ -63,6 +67,55 @@
...
+            logger.warning(f"PID {pid}: Failed to set CPU priority: {e}")
+
+    # Set I/O priority (ionice) - Linux only
+    if io_priority is not None:
```
HAOCHENYE: ionice only affects local block I/O; no-op on network/parallel FS (NFS, GPFS, Lustre). Please check its necessity

xtuner/v1/model/base.py
```diff
@@ -1477,6 +1793,196 @@ def _get_safe_tensor_num(self, dtype: torch.dtype) -> int:
...
+            )
+            if tensors:
+                written_files.append(safetensor_name)
+        return written_files
+
+    def _save_hf_safetensors_file(self, tensors: dict[str, torch.Tensor], filename: Path, hf_dir: Path) -> None:
+        if not tensors:
+            return
+
+        lock_slots = self.config.hf_save_cfg.writer_save_file_lock_slots
```
HAOCHENYE: Is it necessary to control the number of concurrent saving processes here? Under what circumstances would it conflict with training resources?

It feels like using file locks in distributed storage can be a bit unstable; we've had issues with cpfs and lustre in the past.

xtuner/v1/train/trainer.py
```diff
@@ -1639,18 +1661,64 @@ def _maybe_save_hf(self):
             return
 
         save_hf_path = self.exp_dir / f"hf-{self.cur_step}"
+        if self._async_hf_export:
+            async_cleanup_hf_dirs: list[str | Path] = []
+            if self._pending_async_hf_handle is not None:
+                assert self._pending_async_hf_meta is not None
```
HAOCHENYE: For this part, refer to `async_dcp_save`. For the trainer, async_hf simply returns a future that requires waiting

xtuner/v1/model/base.py
```diff
@@ -92,7 +160,14 @@ class HFSaveCfg(PydanticBaseModel):
     model_config = ConfigDict(extra="forbid")
     worker_per_rank: Annotated[int, Parameter(group="model")] = 16
     max_save_rank: Annotated[int, Parameter(group="model")] = 16
+    # Max bytes per generated safetensors shard.
     bucket_size: Annotated[int, Parameter(group="model")] = 1024**3 * 4
+    # Lower async HF writer CPU scheduling priority. 19 is the lowest nice priority for normal users.
+    cpu_priority: Annotated[int, Parameter(group="model")] = 19
+    # Lower async HF writer I/O priority. Linux ionice class 3 means idle I/O priority.
+    io_priority: Annotated[Optional[int], Parameter(group="model")] = 3
```
HAOCHENYE: Align the async_dcp_save, use the same environment variables, and no longer set parameters independently

xtuner/v1/model/base.py
```diff
@@ -662,6 +739,241 @@ def save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16,
...
+            hf_dir = Path(hf_dir)
+        tmp_hf_dir = hf_dir.with_name(f"{hf_dir.name}.incomplete")
+        if rank == 0:
+            if tmp_hf_dir.exists():
+                rmtree(tmp_hf_dir)
+        if dist.is_initialized():
+            dist.barrier()
+        tmp_hf_dir.mkdir(parents=True, exist_ok=True)
+        if dist.is_initialized():
+            dist.barrier()
```
HAOCHENYE: barrier is not necessary, and only mkdir in rank0

xtuner/v1/model/base.py
caixianzhang: Done. Moved `_set_process_qos` to `xtuner/v1/utils/process.py` and
  updated async HF call sites to use `set_process_qos`.

xtuner/v1/model/base.py
caixianzhang: Done. `PermissionError` is covered by `OSError`, so I simplified the
  CPU priority exception handler to `except OSError`.

  I also updated the I/O priority handler to catch
  `subprocess.CalledProcessError` and `OSError`, which covers
  `FileNotFoundError` and `PermissionError`.

xtuner/v1/model/base.py
caixianzhang: This is only used to avoid too many concurrent async HF file writes.
  The lock is acquired only during the actual safetensors write section,
  and once the HF directory has been generated, it will no longer be
  accessed. It is not part of correctness or long-lived coordination, so
  the risk should be limited.

  The option is configurable and can remain disabled on filesystems where
  file locking is known to be unreliable, such as some CPFS/Lustre
  deployments.

xtuner/v1/train/trainer.py
caixianzhang: Done. I aligned async HF with the async DCP save pattern.

  Now the trainer only keeps a pending async HF handle and waits for it
  before launching the next async HF export / at train end. The previous
  trainer-side async HF meta dict was removed, and the wait/finalize
  logic is centralized in `_wait_for_pending_async_hf()`.

  The async HF commit and status checks remain inside `wait_async_hf()`,
  so the trainer no longer tracks async HF internal details.

xtuner/v1/model/base.py
caixianzhang: Done. I removed these async HF runtime tuning knobs from `HFSaveCfg`
  and aligned them with async DCP style.

  Now they are controlled by environment variables instead of model
  config parameters:
  - `XTUNER_ASYNC_SAVE_CPU_PRIORITY`, default `19`
  - `XTUNER_ASYNC_SAVE_IO_PRIORITY`, default `3`
  - `ASYNC_DCP_FILE_WRITE_LOCK_SLOTS`, default `1`

  `ASYNC_DCP_FILE_WRITE_LOCK_SLOTS` is reused for async HF file-write
  throttling as well, so async HF and async DCP share the same file-lock
  concurrency control env var.

xtuner/v1/model/base.py
caixianzhang: Done. I removed the extra barriers around the temporary HF directory
  creation and now only rank 0 removes/recreates the `.incomplete`
  directory.

  The following async HF preparation step already involves distributed
  parameter collection, so we do not need a dedicated barrier just for
  mkdir.

xtuner/v1/train/trainer.py
```diff
@@ -1609,6 +1621,148 @@ def _maybe_save_hf(self):
...
+
+        self._pending_async_hf = {
+            "pid": pid,
+            "step": self.cur_step,
+            "epoch": self._cur_epoch,
+            "save_hf_path": save_hf_path,
+            "tmp_hf_path": tmp_hf_path,
+            "latest_hf_link": latest_hf_link,
+            "meta_path": meta_path,
+            "async_writer_status_path": self.exp_dir
```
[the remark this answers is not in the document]
caixianzhang: This has been addressed in the current implementation.
  `wait_async_hf()` now removes the per-rank status file and cleanup
  marker after a successful global consistency check.

HAOCHENYE (approved):

PR #1694, opened 2026-04-21
[Refactor] Refactor load spec save planning

HAOCHENYE:
## Summary

This PR continues the `LoadSpec` refactor and moves HF load/save layout planning behind clearer plan objects and helpers.

## Changes

- Add `LoadSpec.from_tensor(...)` so DTensor + HF key metadata can be converted into a runtime `LoadSpec` through one entry point.
- Keep `LoadSpec` focused on same-dtype runtime tensor <-> safetensors layout mapping, with fp8 runtime details handled outside the schema.
- Introduce HF save planning and unshard scheduling in `xtuner/v1/utils/load_spec.py`, so `BaseModel` no longer needs to understand preserved fused shard indices directly.
- Simplify `HFSavePlan`: it exposes the concrete `hf_keys` covered by the save tensor plus `preserves_shards`, instead of nesting a load plan or exposing separate global/local key concepts.
- Add targeted unit coverage for deriving `LoadSpec` from plain tensors/DTensors, preserved-shard save key selection, and the save unshard scheduler batching/serialization behavior.

## RL Weight Sync

- Restore the RL EP sequential update behavior so large fused MoE weights can be pushed bucket-by-bucket instead of materializing and submitting the full fused tensor on every EP rank.

## Cleanup

- Remove legacy/commented update code and the old dim-0 no-allocation fuse helper that no longer matched the batched foreach allgather scheduler.

### DIFF
--- a/docs/design/load_spec_refactor.md
+++ b/docs/design/load_spec_refactor.md
@@ -1,0 +1,403 @@
+# LoadSpec 设计
+
+> 面向 `xtuner/v1/utils/load_spec.py` 与 `xtuner/v1/model/base.py` 的加载/保存路径。
+> TP 设计（`dense_tp.md`）依赖本文档描述的抽象。
+
+## TL;DR
+
+LoadSpec 描述 xtuner 运行时 tensor 与 HF safetensors 之间的**纯布局映射**。对一个
+param，它回答两件事：
+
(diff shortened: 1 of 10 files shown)


HAOCHENYE:
@claude review

PR #1703, opened 2026-04-22
Add async checkpoint feature

VincentCheungKokomo:
- feat(checkpoint): add async DCP checkpoint saving
1. Use PyTorch DCP async_save to offload checkpoint file writes to a background process, allowing training to continue while checkpoint I/O is in progress.
2. Create a dedicated gloo process group for async checkpoint collectives to avoid interfering with the training communication group.
3. Write async checkpoints to an .incomplete directory first, then rename after all ranks finish, so partially written checkpoints are never exposed as valid checkpoints.
4. Keep at most one async checkpoint in flight and wait for pending saves before the next checkpoint or trainer shutdown.
5. Reuse a shared-memory staging cache in XtunerCacheWriter to reduce CPU buffer allocation and process handoff overhead between checkpoints.
6. Add optional file write lock slots to throttle concurrent DCP file writes and reduce storage contention during async checkpointing.
7. Store model and optimizer DCP state under a single weights directory and update trainer, RL worker, resume logic, and tests for the merged layout.


xtuner/v1/engine/train_engine.py
```diff
@@ -37,6 +50,53 @@
 from xtuner.v1.utils.grad_norm import cal_grad_norm
 
 
+if BlockingAsyncStager is not None:
```
HAOCHENYE:
```console
In [2]: fw = FileSystemWriter("./")

In [3]: from torch.distributed.checkpoint.staging import AsyncStager, BlockingAsyncStager

In [4]: isinstance(fw, AsyncStager)
Out[4]: True

```

is `_CachingStagingWriter` necessary?

xtuner/v1/engine/train_engine.py
```diff
@@ -387,6 +535,60 @@ def load_dcp(
                     options=_set_options,
                 )
 
+    def load_dcp_merged(
```
HAOCHENYE: The state dict format should be consistant with `async_save` and `save`. If `merged_state_dict` performs better, just replace the current implementation.

xtuner/v1/train/trainer.py
```diff
@@ -522,6 +537,10 @@ def __init__(
 
         self._checkpoint_interval = checkpoint_interval
         self._checkpoint_maxkeep = checkpoint_maxkeep
+        self._async_checkpoint = async_checkpoint
+        self._pending_staging_futures: list[Future] | None = None
+        self._pending_upload_futures: list[Future] | None = None
+        self._pending_checkpoint_finalize: _CheckpointFinalize | None = None
```
HAOCHENYE: Following dcp.async_save, the async interface should return an awaitable future. We can assume there is at most one in-flight async save future in the trainer at any time, and the trainer will always wait for the previous async save to finish before issuing a new one.

xtuner/v1/train/trainer.py
```diff
@@ -783,20 +806,28 @@ def fit(self):
             self._lr_scheduler.step()
             self._maybe_check_health()
             self._maybe_save_hf()
+
+            time_before_checkpoint = time.time()
             ckpt_saved = self._maybe_save(is_snapshot=False)
             if not ckpt_saved:
                 _ = self._maybe_save(is_snapshot=True)
+            checkpoint_time = time.time() - time_before_checkpoint
```
HAOCHENYE: Just log the checkpoint time in train_engine

xtuner/v1/engine/train_engine.py
```diff
@@ -143,6 +147,10 @@ def __init__(
         self.intra_layer_micro_batch = intra_layer_micro_batch
         self._count = 0
         self.has_freeze_params = self.__has_freeze_params()
+        self._async_checkpoint_pg: dist.ProcessGroup | None = None
+        self._async_state_dict_cache: dict[str, Any] | None = None
+        if async_checkpoint:
+            self._async_checkpoint_pg = dist.new_group(backend="gloo")
```
HAOCHENYE: Please leave a comment to describe why we need a gloo process group here.

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +302,133 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+    def async_save_dcp(
+        self,
+        weights_dir: Path,
+        save_optimizer: bool = True,
+    ) -> Future:
+        if not hasattr(dcp, "async_save"):
+            raise RuntimeError(
+                "dcp.async_save is not available in this PyTorch version. "
+                "Please upgrade PyTorch or set async_checkpoint=False."
+            )
```
HAOCHENYE: unnecessary check.

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +302,133 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+
+        if dist.get_rank() == 0:
+            weights_dir.mkdir(parents=True, exist_ok=True)
+
+        state_dict = self._get_dcp_state_dict(cpu_offload=False, save_optimizer=save_optimizer)
+        # Invalidate pinned buffer cache if state_dict structure differs from last call
+        if self._async_state_dict_cache is not None:
+            cached_has_optim = "optimizer" in self._async_state_dict_cache
+            if cached_has_optim != save_optimizer:
+                self._async_state_dict_cache = None
```
HAOCHENYE: when will this branch be triggered?

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +302,133 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+        storage_writer = FileSystemWriter(weights_dir, cache_staged_state_dict=True)
+        storage_writer.state_dict_cache = self._async_state_dict_cache
```
HAOCHENYE: Is this injection necessary?

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +302,133 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+
+        self._async_state_dict_cache = storage_writer.state_dict_cache
+        return future
+
+    def destroy_async_checkpoint_pg(self) -> None:
+        """Destroy the dedicated gloo process group used for async checkpoint."""
+        self._async_state_dict_cache = None
+        if self._async_checkpoint_pg is not None:
+            dist.destroy_process_group(self._async_checkpoint_pg)
+            self._async_checkpoint_pg = None
```
HAOCHENYE, line 483: call it in `__del__`

xtuner/v1/train/trainer.py
```diff
@@ -1109,28 +1229,53 @@ def _maybe_save(self, is_snapshot: bool = False) -> bool:
...
-        )
+        save_optimizer = self._checkpoint_save_optimizer
+        dcp_label = "save_dcp"
+        if self._async_checkpoint and not is_snapshot:
+            t_dcp = time.time()
+            dcp_label = "async_save_dcp"
+            future = self._engine.async_save_dcp(weights_dir=weights_path, save_optimizer=save_optimizer)
+            t_dcp = time.time() - t_dcp
+            # Defer metadata save until async save completes.
+            self._pending_checkpoint = _PendingCheckpoint(
```
HAOCHENYE: Trainer shouldn't need to know about _CheckpointFinalize. Instead, you can call Future.add_done_callback in TrainEngine so that the future tracks timing correctly. The trainer just needs to wait for it (or await it).

xtuner/v1/engine/train_engine.py
VincentCheungKokomo: cache_staged_state_dict keeps pinned staging buffers on the FileSystemWriter instance. XTuner creates one writer per checkpoint path, so carry the cache across writers to preserve steady-state async_save launch performance.

xtuner/v1/train/trainer.py
```diff
@@ -1109,28 +1219,59 @@ def _maybe_save(self, is_snapshot: bool = False) -> bool:
...
+            t_dcp = time.time()
+            dcp_label = "async_save_dcp"
+            future = self._engine.async_save_dcp(weights_dir=weights_path, save_optimizer=save_optimizer)
+            t_dcp = time.time() - t_dcp
+            # Defer metadata save until async save completes.
+            cur_step = self.cur_step
+            cur_epoch = self._cur_epoch
+            train_time_offset = self._train_time + self._train_time_offset
+
+            def finalize_checkpoint_metadata() -> None:
```
HAOCHENYE: Considering keeping the original implementation, I think even if dcp hasn't finished saving, it should be fine to save the meta information first. Try to avoid increasing code complexity just to introduce the asynchronous save feature.

xtuner/v1/train/trainer.py
```diff
@@ -1109,28 +1219,59 @@ def _maybe_save(self, is_snapshot: bool = False) -> bool:
...
-            model_dir=model_path,
-            optimizer_dir=optimizer_path,
-        )
+        save_optimizer = self._checkpoint_save_optimizer
+        dcp_label = "save_dcp"
+        if self._async_checkpoint and not is_snapshot:
+            t_dcp = time.time()
+            dcp_label = "async_save_dcp"
+            future = self._engine.async_save_dcp(weights_dir=weights_path, save_optimizer=save_optimizer)
+            t_dcp = time.time() - t_dcp
```
HAOCHENYE: The time of asynchronous saving should be collected and printed by the engine

autotest/config/qwen3_moe_30BA3_ep8.py
```diff
@@ -47,9 +48,10 @@
     loss_cfg=loss_cfg,
     tokenizer_path=QWEN3_MOE_PATH,
     global_batch_size=16,
-    total_epoch=1,
+    total_step=int(os.environ.get("TOTAL_STEPS", "40")),
```
HAOCHENYE: Revert it back

xtuner/v1/train/trainer.py
```diff
@@ -397,8 +337,10 @@ class TrainerConfig(BaseModel):
     strict_load: bool = True
     checkpoint_interval: int | None = -1
     checkpoint_maxkeep: int | None = -1
+    checkpoint_save_optimizer: bool = True
```
HAOCHENYE: @hhaAndroid Please evaluate whether we actually have a need to not store the optimizer

xtuner/v1/train/trainer.py
```diff
@@ -697,6 +643,7 @@ def __init__(
             load_checkpoint_path=self._load_checkpoint_cfg.checkpoint_path,
             strict=strict_load,
             intra_layer_micro_batch=intra_layer_micro_batch,
+            async_checkpoint=self._async_checkpoint,
```
HAOCHENYE: Why does the build engine need the async flag? Can it be implemented so that if it's asynchronous, it calls engine.async_save, and the engine doesn't need to perceive asynchronous synchronization during the build process?

xtuner/v1/train/trainer.py
```diff
@@ -1155,6 +1113,83 @@ def _maybe_check_health(self):
...
+        future = self._pending_checkpoint
+        self._pending_checkpoint = None
+
+        try:
+            future.result(timeout=timeout)
+        except TimeoutError:
+            future.cancel()
+            raise TimeoutError(f"Async checkpoint timed out after {timeout}s")
+
+    def _finalize_checkpoint_metadata(
```
HAOCHENYE: Keep the original logic unchanged, and do not introduce unnecessary changes temporarily. After obtaining the future, directly execute the subsequent logic.

run_train.sh
```diff
@@ -0,0 +1,361 @@
+#!/bin/bash
```
HAOCHENYE: Revert it

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
-                    state_dict=shard_optimizer_state_dict,
-                    checkpoint_id=optimizer_dir,
-                )
-                if not load_states:
-                    logger.info("Not loading optimizer states")
-                    shard_optimizer_state_dict["state"] = {}
-                if not load_args:
-                    logger.info("Not loading arg defaults")
-                    param_groups = self.optimizer.state_dict()["param_groups"]
-                    # Now we only support one param_group. If we want to support different lr for different parameters,
```
HAOCHENYE, line 369: Please keep the original comments.

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+
+    def _build_async_storage_writer(self, weights_dir: Path, *, save_optimizer: bool) -> XtunerCacheWriter:
+        # XtunerCacheWriter.stage() builds the staging cache directly in POSIX
+        # shared memory when ASYNC_DCP_SHARE_MEMORY=1, so ForkingPickler can
+        # transfer tensors to the daemon via fd handles (no copy).
+        # XTuner creates one writer per checkpoint path; carry the cache across
+        # writers via self._async_state_dict_cache to avoid re-allocation.
+        # Keep cache_staged_state_dict=True: disabling it caused repeated staging
+        # and SIGBUS in experiments.
+        share_memory_cache = os.environ.get("ASYNC_DCP_SHARE_MEMORY", "0") not in {"0", "false", "False"}
```
HAOCHENYE: Change the default value to true.

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+        async_checkpointer_type = AsyncCheckpointerType(os.environ.get("ASYNC_CHECKPOINT_TYPE", "process"))
+        if share_memory_cache and async_checkpointer_type == AsyncCheckpointerType.PROCESS:
+            # In async-process + shared-memory mode the staged tensors are already
+            # CPU-resident. Disable DCP's overlapping CPU loader because it creates
+            # CUDA streams in the checkpoint daemon and can OOM the training GPUs.
+            per_thread_copy_ahead = 0
+        else:
+            per_thread_copy_ahead = int(os.environ.get("ASYNC_DCP_PER_THREAD_COPY_AHEAD", "10000000"))
```
HAOCHENYE: Too many environment variables. We still need to make choices for the users and simplify as much as possible.

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+    def _build_async_storage_writer(self, weights_dir: Path, *, save_optimizer: bool) -> XtunerCacheWriter:
+        # XtunerCacheWriter.stage() builds the staging cache directly in POSIX
+        # shared memory when ASYNC_DCP_SHARE_MEMORY=1, so ForkingPickler can
+        # transfer tensors to the daemon via fd handles (no copy).
+        # XTuner creates one writer per checkpoint path; carry the cache across
+        # writers via self._async_state_dict_cache to avoid re-allocation.
+        # Keep cache_staged_state_dict=True: disabling it caused repeated staging
+        # and SIGBUS in experiments.
+        share_memory_cache = os.environ.get("ASYNC_DCP_SHARE_MEMORY", "0") not in {"0", "false", "False"}
+        async_checkpointer_type = AsyncCheckpointerType(os.environ.get("ASYNC_CHECKPOINT_TYPE", "process"))
```
HAOCHENYE: Just use `process` mode

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+        if share_memory_cache and async_checkpointer_type == AsyncCheckpointerType.PROCESS:
+            # In async-process + shared-memory mode the staged tensors are already
+            # CPU-resident. Disable DCP's overlapping CPU loader because it creates
+            # CUDA streams in the checkpoint daemon and can OOM the training GPUs.
+            per_thread_copy_ahead = 0
+        else:
+            per_thread_copy_ahead = int(os.environ.get("ASYNC_DCP_PER_THREAD_COPY_AHEAD", "10000000"))
+        thread_count = int(os.environ.get("ASYNC_DCP_THREAD_COUNT", "1"))
+        sync_files = os.environ.get("ASYNC_DCP_SYNC_FILES", "1") not in {"0", "false", "False"}
```
HAOCHENYE: For the functionality of async dcp, try to leave only 1-2 environment variable for users to adjust.

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+
+        if self._async_state_dict_cache is not None:
+            cached_has_optim = "optimizer" in self._async_state_dict_cache
```
HAOCHENYE: When will this branch be entered?

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+            logger.info(
+                "[DCP async_save] XtunerCacheWriter "
+                "cache_staged_state_dict=True, "
+                f"per_thread_copy_ahead={per_thread_copy_ahead}, "
+                f"thread_count={thread_count}, sync_files={sync_files}"
+            )
+
+        storage_writer = XtunerCacheWriter(
+            weights_dir,
+            cache_staged_state_dict=True,
```
HAOCHENYE: https://github.com/InternLM/xtuner/pull/1703/changes#diff-6bf3a94613cc6d16c24d59aa79ee24b2aa4d1c45ed19523186bcf74937bd43adR432

Is there a conflict here?

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+        #
+        # XtunerCacheWriter.stage() creates its staging cache directly in POSIX
+        # shared memory (/dev/shm) when ASYNC_DCP_SHARE_MEMORY=1.  PyTorch's
+        # ForkingPickler detects shared-memory tensors and sends fd handles (no
+        # data copy) to the daemon → PSS is split 1/2 each, no amplification.
+        # The shm cache is reused across checkpoints via self._async_state_dict_cache.
+        state_dict = self._get_dcp_state_dict(cpu_offload=False, save_optimizer=save_optimizer)
+        storage_writer = self._build_async_storage_writer(incomplete_dir, save_optimizer=save_optimizer)
+
+        async_checkpointer_type = AsyncCheckpointerType(os.environ.get("ASYNC_CHECKPOINT_TYPE", "process"))
```
HAOCHENYE: https://github.com/InternLM/xtuner/pull/1703/changes#diff-6bf3a94613cc6d16c24d59aa79ee24b2aa4d1c45ed19523186bcf74937bd43adR433

Repeated checks, and using process by default can work

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+                    done_future.result()
+                    if self._async_checkpoint_pg is not None:
+                        dist.barrier(group=self._async_checkpoint_pg)
+                    if dist.get_rank() == 0:
+                        if weights_dir.exists():
+                            shutil.rmtree(weights_dir)
+                        incomplete_dir.rename(weights_dir)
+                    if self._async_checkpoint_pg is not None:
+                        dist.barrier(group=self._async_checkpoint_pg)
+                except BaseException as exc:
```
HAOCHENYE, line 423: This is not the way to do it. An error should be reported if the checkpoint save fails, and the task should not continue training. Otherwise, it will cause more losses.

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+            )
+
+        committed_future: Future = Future()
+
+        def commit_async_save(done_future: Future) -> None:
+            def commit() -> None:
+                try:
+                    done_future.result()
+                    if self._async_checkpoint_pg is not None:
+                        dist.barrier(group=self._async_checkpoint_pg)
```
HAOCHENYE: why `self._async_checkpoint_pg` could be None here

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+                    if dist.get_rank() == 0:
+                        if weights_dir.exists():
+                            shutil.rmtree(weights_dir)
```
HAOCHENYE: First, if weights_dir exists, it should be deleted when entering async_save_hf, not waiting until the end. Second, why does it already exist? If it already exists, does that mean this is a bug that needs to raise an error?

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+                        incomplete_dir.rename(weights_dir)
+                    if self._async_checkpoint_pg is not None:
+                        dist.barrier(group=self._async_checkpoint_pg)
```
HAOCHENYE: Avoid such overly defensive programming

xtuner/v1/engine/train_engine.py
```diff
@@ -294,98 +309,224 @@ def save_hf(self, hf_dir: str, save_dtype: torch.dtype = torch.bfloat16):
...
+                self._async_state_dict_cache = storage_writer.state_dict_cache
+
+                elapsed = time.time() - t0
+                logger.info(f"[DCP async_save for {weights_dir}] finished in {elapsed:.2f}s")
+                if not committed_future.done():
+                    committed_future.set_result(None)
+
+            threading.Thread(target=commit, daemon=True).start()
+
+        dcp_future.add_done_callback(commit_async_save)
```
HAOCHENYE: Where is the purpose of using a committed future + dcp future here? I understand that committed simply involves recording the time consumed, printing some logs, and finally moving the path, which are all lightweight operations. What is the purpose of such an implementation?

xtuner/v1/patch/xtuner_storage.py
```diff
@@ -87,12 +180,69 @@ def __init__(
         self._enable_write_result_caching = enable_write_result_caching
         self._cached_write_results_key = cache_key_prefix + self.__class__.__name__
 
+    def stage(self, state_dict: dict[str, Any]) -> dict[str, Any]:
```
HAOCHENYE, line 181: Can we remove unnecessary environment variables and directly reuse PyTorch's implementation?

xtuner/v1/patch/xtuner_storage.py
```diff
@@ -1,10 +1,16 @@
+import fcntl
```
HAOCHENYE, line 1: Is it necessary to rewrite `FileSystemWriter`? I understand that removing unnecessary environment variables, the core requirement is to lock the file stream. I looked at PyTorch's code, are you implementing a custom `FileSystem` class, which is enough to meet the needs?

xtuner/v1/engine/train_engine.py
HAOCHENYE: Could we hold the storage writer all the time if async_save is enabled?

xtuner/v1/patch/xtuner_storage.py
VincentCheungKokomo: Addressed. The file-locking logic has been moved to a custom FileSystem implementation. XtunerCacheWriter.write_data() now directly reuses PyTorch's FileSystemWriter.write_data() via super().write_data(...), and the lock is only applied in FileSystem.create_stream() when opening writable streams.

I also removed the extra file-lock environment variables. The lock key is derived from the run path, the lock directory is fixed internally, and only ASYNC_DCP_FILE_WRITE_LOCK_SLOTS remains as the write-concurrency tuning knob.

xtuner/v1/patch/xtuner_storage.py
VincentCheungKokomo: Addressed. The env switches were removed and shared-memory staging is now the default. We still keep a small stage() override because PyTorch's default stager creates a pinned CPU cache, while async process checkpointing needs a reusable shared-memory cache to avoid extra inter-process tensor sharing/copy overhead.

HAOCHENYE (approved):

PR #1716, opened 2026-04-24
[Feature] support ep and enhance step latency for muon

nil0x9:
# Summary
Highlight:
- Support Muon training when EP > 1
- reduce latency and memory footprint caused by excessive padding for all2all.

Details:
- Remove distributed_mesh parameter; extract device mesh and process group directly from DTensor metadata to support heterogeneous meshes (ViT and LM have different meshes).
- Pre-compute adjust_lr ratios in `__init__` based on global (unsharded) shape, avoiding incorrect shape references inside async tasks after communication.
- Add MoE expert-parallel support: per-expert Newton-Schulz orthogonalization, requires n_experts % ep_size == 0 on the EP dimension. On the FSDP dimension, skip communication when n_experts % fsdp_size == 0 (each rank holds complete experts), use sub-group all-gather when fsdp_size % n_experts == 0, otherwise fall back to batched all-to-all.
- Add AGRS (All-Gather + Reduce-Scatter) path for remainder batches to avoid zero-padding overhead, with even-sharding guard to prevent deadlock.
- Refactor shared utilities (group_tensors_by_device_mesh_and_placements,  cal_total_norm) from grad_norm.py to dtensor.py.
- Remove `# type: ignore` from file head, fix lint, add full type annotations.

## Communication Path Table

| Path | Condition | Communication | Computation | Code Location |
|------|-----------|---------------|-------------|---------------|
| **No communication** | MoE, `n_experts % fsdp_size == 0` (each rank holds complete experts) | None | Local NS per expert | `skip_communication = True` |
| **Sub-group all-gather** | MoE, `fsdp_size % n_experts == 0`, batch < fsdp_size | Small all-gather within sub-group | Local NS on reconstructed expert, then slice back | `subgroup_process_group` path |
| **All-to-all (even)** | Batch fills `world_size`, `global_dim % world_size == 0` | Forward a2a + backward a2a | Each rank orthogonalizes 1 full matrix | Even sharding branch |
| **All-to-all (uneven)** | Batch fills `world_size`, `global_dim % world_size != 0` | Padded forward a2a + padded backward a2a | Each rank orthogonalizes 1 full matrix (after unpad) | Uneven sharding branch |
| **AGRS** | Batch < `world_size`, evenly sharded (`dim % world_size == 0`) | All-gather + reduce-scatter | Selective NS: rank `r` computes params where `i % W == r` | `use_agrs = True` |
| **Padded all-to-all** | Batch < `world_size`, unevenly sharded (`dim % world_size != 0`) | Pad batch with `zeros_like` to `world_size`, then even/uneven a2a | Padded zeros go through NS (wasted computation) | Fallback when `is_evenly_sharded = False` |

## Constraints

- EP dimension: `n_experts % ep_size == 0` is **required** (assertion).
- FSDP dimension: no hard constraint; the three MoE paths (no-comm, sub-group, fall-through to a2a) are optimizations, not requirements.

# End Results
Experiments are run on a model of arch Qwen3.5. #GPU=128
## Regression Test

Loss aligns almost perfectly when `XTUNER_DETERMINISTIC` is turned on.
image

## Step Latency
Profile was collected at step 10 (rank0). Latencies might vary run-to-run.
| VERSION | EP_SIZE | latency |
|------|--------|--------|
|Base|1|  783ms|
|Base|8| Not Supported|
|This PR|1| 236ms|
|This PR|8|232ms|

The original implementation wastes a lot of time creating and transmitting padding tensors. This is wasteful when num_layers < world_size. The following graph showcases this behavior:
image

This is especially the case when the world_size grows. Typically the world_size might be somewhere between 64 and 2048. But the number of layers (which determines how many params of the same shape are batched and all2all'ed) is typically smaller than 100, meaning that for large-scale training, the memory footprint and latency caused by padding tensors grows linearly with world size. In my experiment, I encountered OOM when training a 397B MoE model (60 layers, 128GPU, see below) for this exact reason.

The following is the timeline when running the same experiment using code in this PR:
image

## 397B Model
| VERSION | EP_SIZE | latency |
|------|--------|--------|
|Base|1|  OOM|
|Base|8| Not Supported|
|This PR|1| 491ms|
|This PR|8|1019ms|

# Limitation
**Padding fallback**. When the number of experts satisfies neither n_experts % fsdp_size == 0 nor fsdp_size % n_experts == 0, the code falls back to creating large padding tensors and running all-to-all to distribute
  expert parameters across ranks. This incurs significant memory and latency overhead. In practice, however, such model configurations are uncommon.

**Dispatch efficiency**. The current parameter dispatch follows a two-stage pattern: tensors are first grouped by device mesh and sharding placement, then routed within each group. This grouping-before-dispatching approach can introduce waste. A more general global parameter-to-device dispatcher **could** eliminate this overhead and is left for future work.

### DIFF
--- a/xtuner/v1/config/optim.py
+++ b/xtuner/v1/config/optim.py
@@ -185,14 +185,9 @@ class MuonConfig(OptimConfig):
                     f"Muon params: {num_muon_regular / 1e6:.2f}M, AdamW params: {num_adamw / 1e6:.2f}M (counts by numel)"
                 )
             logger.info(f"Untrainable parameters names: {untrainable_names}")
-            logger.info(
-                f"using Muon optimizer distributed_mesh_size: {model.fsdp_mesh.size()}, "
-                f"distributed_mesh: {model.fsdp_mesh}"
-            )
 
         optimizer = Muon(
             param_groups,
-            distributed_mesh=model.language_model.fsdp_mesh,  # TODO: 暂不支持 EP>1; maybe rm device_mesh dependency?
             lr=self.lr,
             mu=self.momentum,
             betas=self.betas,
--- a/xtuner/v1/optim/muon.py
+++ b/xtuner/v1/optim/muon.py
@@ -1,4 +1,3 @@
-# type: ignore
 # ================================================================
 # Copyright (c) 2026 InternLM. All rights reserved.
 #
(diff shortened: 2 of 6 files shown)


nil0x9:
@claude review

CyCle1024 (approved):

PR #1806, opened 2026-05-18
[Fix] Fix missing post_micro_batch_forward invocation in vl model

nil0x9:
Currently vl base model does not override `post_micro_batch_foward` method, meaning that they inherit the behavior from `BaseModel` directly. When one trains  a vl model with MoE text part, this would cause:
1. missing `maxvio` from tracking logs;
2. (**critical**) if `NoAuxRouter` is used, the bias would never be updated.

This PR fixes this issue.


nil0x9:
@claude review

PR #1829, opened 2026-05-22
[Fix] Fix float8 unpickling error when resume

nil0x9:
PR #1517 (commit a4a91e6) refactored Float8Handler initialization from an eager call in TrainEngine.build_model() to a lazy-initialized property
on BaseModel.

As a side effect, the torch.serialization.add_safe_globals() call inside Float8Handler.__init__ is now only triggered on the first training step instead of at model build time. This call registers WeightWithDynamicTilewiseFloat8CastTensor and
WeightWithDynamicTensorWiseFloat8CastTensor as safe for unpickling.

When resuming from a DCP checkpoint with float8 enabled, load_dcp() calls dcp.load() before any training step has run, so add_safe_globals() has not been called yet. PyTorch's weights_only=True deserialization then rejects the float8 tensor types as unsafe globals, raising _pickle.UnpicklingError.

Fix this by eagerly accessing mod.float8_handler for all BaseModel modules at the top of load_dcp(), ensuring add_safe_globals() is called before dcp.load() regardless of whether a training step has previously run.

### DIFF
--- a/tests/engine/test_moe_train_engine_float8.py
+++ b/tests/engine/test_moe_train_engine_float8.py
@@ -2,6 +2,7 @@ import os
 import tempfile
 import shutil
 import time
+from pathlib import Path
 
 import parametrize
 import torch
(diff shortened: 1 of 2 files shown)


nil0x9:
@claude review

HAOCHENYE (approved):

PR #1843, opened 2026-05-26
[Enhance] Show dataset build progress with cache status

HAOCHENYE:
Add a tqdm progress bar to build_datasets and expose JsonlDataset.cached
so the bar can report whether each dataset hit the tokenize cache. Keep
each jsonl path bound to its own dataset/tokenize_fn config while
expanding paths, instead of reusing the last loop config across all
paths. Also fix a stale get_rank comparison (get_rank == 0) that left the
token-count debug log permanently disabled.


HAOCHENYE:
@claude review

hhaAndroid (approved):

xtuner/v1/model/moe/qwen3_5_text_split.py
```diff
@@ -1,239 +0,0 @@
-import re
```
hhaAndroid, line 1: 这个为啥直接删掉？

PR #1852, opened 2026-05-28
fix distributed barrier imbalance in async_save_dcp

VincentCheungKokomo:
commit d66ff05d59dbfc4ab176125d84c5dd11df02eabd
fix(engine): fix distributed barrier imbalance in async_save_dcp

Three correctness fixes for async_save_dcp on multi-rank training:

1. Add dist.barrier() after rank-0 rmtree+mkdir to prevent non-rank-0
   processes from writing into an incomplete directory while rank-0 is
   still cleaning up a stale .incomplete dir.

2. Use dist.all_reduce(MAX) to broadcast the retry/fatal decision so
   all ranks agree before raising or retrying. Without this, ranks that
   classify the exception differently would deadlock at the barrier.

3. Only query weights_dir.exists() on rank-0 and broadcast the result
   via dist.broadcast, ensuring all ranks raise FileExistsError (or
   proceed) together even under NFS cache inconsistency.

commit 7ee155775ffb1248bb506bba92eefba0c4c3e65e
 [Fix] Fix async DCP checkpoint "received 0 items of ancdata" and add early failure detection

  1. Coalesce per-tensor shared memory into per-dtype buffers to reduce fd count
     from ~3000 to ~2 during daemon subprocess handoff, fixing the ancdata bug.

  2. Add warmup_async_save_dcp() to trigger daemon init before training starts,
     surfacing port conflicts (EADDRINUSE) immediately instead of mid-training.

  3. Add _check_async_save_health() to detect async save failures within one
     step rather than waiting until the next checkpoint interval.

  4. Allow snapshot saves to use async path.

### DIFF
--- a/xtuner/v1/engine/train_engine.py
+++ b/xtuner/v1/engine/train_engine.py
@@ -513,6 +513,51 @@ class TrainEngine:
         storage_writer.state_dict_cache = self._async_state_dict_cache
         return storage_writer
 
+    @classmethod
+    def warmup_async_save_dcp(cls, work_dir: Path) -> None:
+        """Warm up async DCP save infrastructure with a tiny dummy state dict.
+
+        This triggers the full async save path — including daemon subprocess
+        spawn and its internal init_process_group — so that errors like port
+        conflicts (EADDRINUSE) surface before any real training begins.
+
+        Args:
+            work_dir (Path): Working directory for temporary preflight files.
+        """
+        preflight_dir = work_dir / ".preflight_dcp"
+        weights_dir = preflight_dir / "weights"
+
+        if dist.get_rank() == 0:
+            if preflight_dir.exists():
+                shutil.rmtree(preflight_dir)
+            weights_dir.mkdir(parents=True, exist_ok=True)
+        dist.barrier()
+
+        dummy_state_dict = {"_preflight": torch.zeros(1)}
+
+        try:
+            async_save_kwargs: dict[str, Any] = {}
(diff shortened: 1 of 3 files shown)


xtuner/v1/engine/train_engine.py
```diff
@@ -392,12 +392,20 @@ def async_save_dcp(
         # temporary .incomplete directory and commit it only after every rank's
         # async_save future has completed.
         incomplete_dir = weights_dir.with_name(f"{weights_dir.name}.incomplete")
-        if weights_dir.exists():
+        # Check existence only on rank 0 and broadcast the result so all ranks
+        # raise (or continue) together. Without this, NFS cache inconsistencies
+        # could cause some ranks to raise while others proceed to the barrier,
+        # resulting in a deadlock.
+        dir_exists = torch.tensor(int(weights_dir.exists() if dist.get_rank() == 0 else 0), dtype=torch.int32)
```
HAOCHENYE, line 399: Can we only check if the directory exists on rank0?

VincentCheungKokomo: This already only calls exists() on rank 0 (via weights_dir.exists() if dist.get_rank() == 0 else 0) — other ranks simply use 0.  
  The next line dist.broadcast(dir_exists, src=0, group=async_checkpoint_pg) then syncs rank 0's result to all ranks. This way all  
  ranks get a consistent decision and either all raise or all proceed together, avoiding a deadlock where NFS cache inconsistencies 
  cause some ranks to raise while others continue to the barrier.

PR #1858, opened 2026-05-31
Refactor async HF metadata prewrite and background finalize

caixianzhang:
## Summary

This PR refactors async HF export so metadata is prepared before the safetensors writer starts, and the final checkpoint directory is published by a background finalize thread as soon as the writer completes.

It supersedes #1857 and carries over its fixes:
- async HF non-weight files now prefer `config.save_hf()` / `hf_config`, matching sync HF behavior, and only fall back to copying from `_hf_path`
- compose async HF writes `model.safetensors.index.json` with the same JSON formatting as compose sync HF

Main changes:
- add async HF runtime resources initialized lazily on the first async HF save
- create a dedicated Gloo process group for async HF control-plane collectives
- prewrite config/tokenizer/processor files and `model.safetensors.index.json` into hidden `.hf-*.incomplete` directories before launching the writer process
- keep the writer process focused on safetensors shard I/O and per-rank status writing
- add a single-worker background finalize executor that joins the writer, gathers per-rank status, and renames `.hf-*.incomplete` to `hf-*` once all ranks succeed
- return `Future[Path]` from async HF save and finalize HF/DCP metadata from future callbacks after the underlying save succeeds
- add `AsyncSaveMonitor` to observe async HF/DCP futures, report background failures, and terminate the training process group
- use async HF-specific file-write lock slots via `ASYNC_HF_FILE_WRITE_LOCK_SLOTS`; async HF no longer reads `ASYNC_DCP_FILE_WRITE_LOCK_SLOTS`
- based on validation feedback, throttle async DCP background file writes by default with `ASYNC_DCP_FILE_WRITE_LOCK_SLOTS=1` to reduce shared-storage and CPU/I/O contention during async saves

## Validation

Passed:
- `ruff check xtuner/v1/utils/async_save_monitor.py xtuner/v1/utils/process.py xtuner/v1/model/base.py xtuner/v1/model/compose/base.py xtuner/v1/engine/train_engine.py xtuner/v1/train/trainer.py tests/utils/test_async_save_monitor.py tests/train/test_trainer.py`
- `ruff format --check xtuner/v1/utils/async_save_monitor.py xtuner/v1/utils/process.py xtuner/v1/model/base.py xtuner/v1/model/compose/base.py xtuner/v1/engine/train_engine.py xtuner/v1/train/trainer.py tests/utils/test_async_save_monitor.py tests/train/test_trainer.py`
- `python3 -m py_compile xtuner/v1/utils/async_save_monitor.py xtuner/v1/utils/process.py xtuner/v1/model/base.py xtuner/v1/model/compose/base.py xtuner/v1/engine/train_engine.py xtuner/v1/train/trainer.py tests/utils/test_async_save_monitor.py tests/train/test_trainer.py`
- `tests/train/test_trainer.py::TestHooksConfig::test_async_hf_save_hook_timing` passed on cluster
- `tests/utils/test_async_save_monitor.py` passed on cluster

Cluster validation:
- `bash sft_files/run_qwen3_moe_async_hf_tests_8gpu.sh`
- `bash sft_files/run_qwen3_vl_moe_async_hf_tests_8gpu.sh`
- `RUN_LABELS="async_hf sync_hf" bash sft_files/run_qwen3_moe_30ba3_ep8_hf_mem_suite.sh`
- `RUN_LABELS="async_hf sync_hf" bash sft_files/run_qwen3_vl_moe_30ba3_ep8_hf_mem_suite.sh`
- `bash sft_files/compare_hf_non_safetensors.sh`
- `bash sft_files/check_hf_from_pretrained.sh`
- `bash sft_files/run_qwen3_moe_async_hf_fail_rank2_hf90.sh`


tests/train/test_trainer_async_hf.py
HAOCHENYE, line 1: Delete this fake test and directly run a real asynchronous test at https://github.com/InternLM/xtuner/blob/d92529d6ec9de1d57fc8919878666e9bf16e3ba5/tests/train/test_trainer.py#L485.

In addition to end-to-end correctness, we can use hooks to verify that in the asynchronous scenario, the hf and dcp saves at the N-th step are completed in a timely manner without lagging behind.

xtuner/v1/utils/async_save_monitor.py
```diff
@@ -0,0 +1,85 @@
...
+                    break
+                if exc is not None:
+                    failure = (item, exc)
+                    break
+
+        if failure is not None:
+            self._terminate_failure(failure)
+
+    def _terminate_failure(self, failure: tuple[AsyncSaveWatchItem, BaseException]) -> None:
+        with self._lock:
```
HAOCHENYE: Unnecessary locks?

xtuner/v1/model/compose/base.py
```diff
@@ -171,40 +173,55 @@ def save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16,
...
+        elif self._hf_path is not None:
+            for file in cast(Path, self._hf_path).iterdir():
+                if file.suffix != ".safetensors":
+                    target_path = hf_dir / file.name
+                    if file.is_file():
+                        copy(file, target_path)
+                    else:
+                        copytree(file, target_path, ignore_dangling_symlinks=True, dirs_exist_ok=True)
+        else:
+            raise RuntimeError("Internal Error, both self.config.hf_config and self._hf_path are None")
```
HAOCHENYE: Extract this part of the code into model base.py for shared use.

xtuner/v1/model/compose/base.py
```diff
@@ -171,40 +173,55 @@ def save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16,
...
+
+        with open(hf_dir / "model.safetensors.index.json", "w") as f:
+            json.dump({"weight_map": dict(weight_map), "metadata": {}}, f, indent=4)
+
+    def destroy_async_hf_resources(self) -> None:
+        super().destroy_async_hf_resources()
+        for module in (self.language_model, self.vision_tower, self.multi_modal_projector):
+            module._async_hf_tensor_cache.clear()
```
HAOCHENYE: 同理，在 base 里可以用 isinstance 去遍历所有的 BaseModel，去执行 `_async_hf_tensor_cache`

xtuner/v1/model/base.py
```diff
@@ -684,46 +687,95 @@ def save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16,
         with profile_time_and_memory(f"[Saving HF to [{safetensors_prefix}]{hf_dir} cost]"):
             self._save_hf(hf_dir=hf_dir, save_dtype=save_dtype, safetensors_prefix=safetensors_prefix)
 
+    def init_async_hf_resources(self) -> None:
```
HAOCHENYE: Should not be a public interface

xtuner/v1/model/base.py
```diff
@@ -96,8 +95,11 @@ class AsyncHFSaveHandle:
     process: Any
     hf_dir: Path
     tmp_hf_dir: Path
-    status_path: Path
-    cleanup_done_path: Path
+
+
+@dataclass
+class AsyncHFResources:
+    finalize_pg: dist.ProcessGroup | None
```
HAOCHENYE, line 106: Unnecessary single-member abstraction

xtuner/v1/model/base.py
```diff
@@ -684,46 +687,95 @@ def save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16,
         with profile_time_and_memory(f"[Saving HF to [{safetensors_prefix}]{hf_dir} cost]"):
             self._save_hf(hf_dir=hf_dir, save_dtype=save_dtype, safetensors_prefix=safetensors_prefix)
 
+    def init_async_hf_resources(self) -> None:
+        if self._async_hf_resources is not None:
```
HAOCHENYE, line 704: https://github.com/InternLM/xtuner/pull/1858/changes#diff-d67fb488dc8f8ed52b7ee18b4f1e52f870d1cdff28d38a406b2bea0b4c70bb97R723

Same as here

xtuner/v1/model/base.py
```diff
@@ -684,46 +687,95 @@ def save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16,
...
+        if not dist.is_initialized():
+            return [local_obj]
+
+        gathered: list[Any] = [None for _ in range(dist.get_world_size())]
+        dist.all_gather_object(gathered, local_obj, group=self._get_async_hf_finalize_pg())
+        return gathered
+
+    def _barrier_async_hf(self) -> None:
+        if dist.is_initialized():
+            dist.barrier(group=self._get_async_hf_finalize_pg())
```
HAOCHENYE: Remove the check, as xtuner's code already depends on distributed initialization by default

xtuner/v1/model/base.py
```diff
@@ -814,24 +858,21 @@ def _prepare_async_hf_snapshot(
...
-    def wait_async_hf(self, handle: AsyncHFSaveHandle | None = None) -> Path | None:
-        if handle is None:
-            handle = self._pending_async_hf
-        if handle is None:
-            return None
+    def _clear_pending_async_hf(self, handle: AsyncHFSaveHandle) -> None:
+        if self._pending_async_hf is handle:
+            self._pending_async_hf = None
 
+    def commit_async_hf_save(
```
HAOCHENYE: private internface

xtuner/v1/model/base.py
```diff
@@ -684,46 +687,95 @@ def save_hf(self, hf_dir: Path | str, save_dtype: torch.dtype = torch.bfloat16,
...
+            dist.destroy_process_group(resources.finalize_pg)
+
+    def _get_async_hf_resources(self) -> AsyncHFResources:
+        if self._async_hf_resources is None:
+            self.init_async_hf_resources()
+        assert self._async_hf_resources is not None
+        return self._async_hf_resources
+
+    def _get_async_hf_finalize_pg(self) -> dist.ProcessGroup | None:
+        return self._get_async_hf_resources().finalize_pg
```
HAOCHENYE: The code here is too redundant

xtuner/v1/model/compose/base.py
```diff
@@ -244,41 +263,41 @@ def async_save_hf(
             process=process,
             hf_dir=hf_dir,
             tmp_hf_dir=tmp_hf_dir,
-            status_path=status_path,
-            cleanup_done_path=cleanup_done_path,
+        )
+        commit_executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="async-hf-commit")
+        commit_future = commit_executor.submit(
+            self.commit_async_hf_save,
+            handle,
```
HAOCHENYE: The thread pool itself can be persisted; just call submit each time you save a checkpoint

tests/train/test_trainer_async_hf.py
```diff
@@ -104,23 +101,28 @@ def async_save_hf(self, hf_dir: Path | str, cleanup_hf_dirs=()):
...
+            future.set_result(self.commit_async_hf_save(payload))
+        except BaseException as exc:
+            future.set_exception(exc)
+        if future.done():
+            self._clear_pending_async_hf(future)
+        return future
+
+    def _clear_pending_async_hf(self, future):
+        if self._pending_async_hf is future:
+            self._pending_async_hf = None
```
HAOCHENYE: What is the reason the model needs to maintain the pending HF state? Is it to prevent contamination of the staging tensor when the next one arrives before the previous one has finished saving? Some comments are needed here

xtuner/v1/model/base.py
```diff
@@ -748,42 +795,39 @@ def async_save_hf(
...
+
+        def clear_pending_async_hf(_: Future[Path]) -> None:
+            self._clear_pending_async_hf(handle)
+
+        def shutdown_commit_executor(_: Future[Path]) -> None:
+            commit_executor.shutdown(wait=False)
+
+        commit_future.add_done_callback(clear_pending_async_hf)
+        commit_future.add_done_callback(shutdown_commit_executor)
+        return commit_future
```
HAOCHENYE, line 822: The code here feels a bit redundant. If we don't manually craft a Process and directly use cocurrent's PoolExecutor, then the callback will be executed in the built-in thread monitor by default, and this part of our code can be simplified.

However, considering that the changes would be quite large, we can temporarily keep this writing style for now, and try ProcessExecutor later to maintain processes.

xtuner/v1/model/base.py
HAOCHENYE: If you want to keep this abstraction, I suggest including the thread executor as well

xtuner/v1/model/base.py
```diff
@@ -840,89 +881,26 @@ def wait_async_hf(self, handle: AsyncHFSaveHandle | None = None) -> Path | None:
         elif exit_code != 0:
             local_ok = False
             local_error = f"child_exit_code={exit_code}"
```
HAOCHENYE: Is a timeout needed? Could a process get stuck and never exit, causing the entire training to freeze?

xtuner/v1/model/base.py
```diff
@@ -840,89 +881,26 @@ def wait_async_hf(self, handle: AsyncHFSaveHandle | None = None) -> Path | None:
...
                 rmtree(hf_dir)
-            tmp_hf_dir.rename(hf_dir)
-        if dist.is_initialized():
-            dist.barrier()
-        if self._pending_async_hf is handle:
-            self._pending_async_hf = None
+            handle.tmp_hf_dir.rename(hf_dir)
+        self._barrier_async_hf()
+        log_rank0.info(f"[Async saving HF to {hf_dir}] finalized")
+        self._barrier_async_hf()
```
HAOCHENYE: too much barrier

xtuner/v1/train/trainer.py
```diff
@@ -604,9 +606,10 @@ def __init__(
         self._checkpoint_interval = checkpoint_interval
         self._checkpoint_maxkeep = checkpoint_maxkeep
         self._async_hf_export = async_hf_export
-        self._pending_async_hf_handle: AsyncHFSaveHandle | None = None
-        self._pending_async_hf_step: int | None = None
-        self._pending_async_hf_epoch: int | None = None
+        self._pending_async_hf_future: Future[Path] | None = None
+        self._pending_async_hf_finalize_done: threading.Event | None = None
+        self._async_save_monitor = AsyncSaveMonitor()
+        self._save_finalize_lock = threading.RLock()
```
HAOCHENYE: Add an internal function to initialize asynchronous related resources

xtuner/v1/train/trainer.py
```diff
@@ -884,10 +890,14 @@ def fit(self):
...
 
-        self._wait_for_pending_async_hf()
+        if self._async_hf_export:
+            self._wait_for_pending_async_hf()
+            self._engine.model.destroy_async_hf_resources()
 
         # TODO: Should use flush rather than close
         self._wait_for_pending_checkpoint()
+        if self._async_hf_export or self._async_checkpoint:
+            self._async_save_monitor.stop()
```
HAOCHENYE, line 907: The style is different. The hf part destroys within the if, while the dcp part destroys uniformly. Although both are acceptable, I prefer to destroy within the if

xtuner/v1/train/trainer.py
```diff
@@ -1190,18 +1199,16 @@ def _maybe_check_health(self):
                 raise RuntimeError("Health check failed, exit training")
             log_rank0.info(f"Health check passed at step {self.cur_step}")
 
-    def _wait_for_pending_checkpoint(self, timeout: int = 3000) -> None:
+    def _wait_for_pending_checkpoint(self) -> None:
```
HAOCHENYE, line 1208: Where is the current timeout handling?

xtuner/v1/train/trainer.py
```diff
@@ -1284,28 +1292,29 @@ def _maybe_save(self, is_snapshot: bool = False) -> bool:
...
+            current_exp.history[-1]["end"] = self.cur_step
+
+            # Delete checkpoints and update meta's checkpoint_list
+            ckp_maxkeep = self._checkpoint_maxkeep if not is_snapshot else 1
+            if ckp_maxkeep is not None and ckp_maxkeep > 0 and len(ckp_list) > ckp_maxkeep:
+                ckp_pop_num = len(ckp_list) - ckp_maxkeep
+                for _ in range(ckp_pop_num):
+                    deleted_ckp = ckp_list.pop(0)
+                    if self.rank == 0 and Path(deleted_ckp).exists():
+                        rmtree(deleted_ckp)
```
HAOCHENYE, line 1373: It needs to be executed via a callback, similar to finalize_hf. Otherwise, DCP may finish saving the meta information but the weights themselves are not fully saved, and if the process terminates midway, it will eventually lead to reading incorrect meta information upon resume.

HAOCHENYE:
@claude review

tests/train/test_trainer_async_hf.py
caixianzhang: Resolved this by removing the fake-engine based async HF test and
  replacing it with a real Trainer path test.

  Changes made:

  - Removed tests/train/test_trainer_async_hf.py.
  - Added tests/train/test_trainer.py::TestHooksConfig::test_async_hf_save_hook_timing.

  - The new test uses real Trainer.from_config() / TrainEngine / BaseModel.async_save_hf() instead of mocking Trainer.build_engine.

  - The test enables async HF with total_step=4, hf_interval=2, and verifies via after_save_hf hook that HF saves are finalized at steps 2 and 4.

  - The hook also checks that the finalized HF directory is visible when the hook runs, and the test verifies
    trainer.meta.latest_exp.hf_checkpoint_list is updated to hf-2 and hf-4.

  Validation result:

  tests/train/
  test_trainer.py::TestHooksConfig::test_async_hf_save_hook_timing PASSED
  1 passed, 5 warnings in 33.87s

caixianzhang: image

xtuner/v1/utils/async_save_monitor.py
caixianzhang: simplify the failure path: `_terminate_failure()` no longer uses `_terminated` or an extra lock, because only the single monitor worker triggers termination and the process group exits immediately after logging the failure.

xtuner/v1/model/compose/base.py
caixianzhang: Removed `BaseComposeModel.destroy_async_hf_resources()`.
 Updated `BaseModel.destroy_async_hf_resources()` to iterate over `self.modules()` and clear `_async_hf_tensor_cache` for every submodule that is a `BaseModel`.

xtuner/v1/model/compose/base.py
caixianzhang: Removed the duplicated `_write_hf_index_and_config()` implementation from `xtuner/v1/model/compose/base.py`.
    `BaseComposeModel` now reuses `BaseModel._write_hf_index_and_config()`.

xtuner/v1/model/base.py
caixianzhang: Resolved by renaming `init_async_hf_resources()` to `_init_async_hf_resources()` and keeping it only as an internal lazy-init path through `_get_async_hf_resources()`.

xtuner/v1/model/base.py
caixianzhang: I kept `AsyncHFResources`, but made it own more than just `finalize_pg`.
It now contains both:
  - `finalize_pg`
  - the async HF `commit_executor`

xtuner/v1/model/base.py
caixianzhang: Addressed. `init_async_hf_resources()` has been renamed to `_init_async_hf_resources()` and is now only called internally by `_get_async_hf_resources()`.

xtuner/v1/model/base.py
caixianzhang: Addressed by renaming `commit_async_hf_save()` to `_commit_async_hf_save()` since it is only used internally by the async HF commit executor.

xtuner/v1/model/base.py
caixianzhang: Addressed by removing `_get_async_hf_finalize_pg()` and directly using `self._get_async_hf_resources().finalize_pg` in the async HF gather/barrier helpers. I also removed the redundant `dist.is_initialized()` checks there.

xtuner/v1/model/compose/base.py
caixianzhang: Addressed. The async HF commit executor is now persisted in `AsyncHFResources` and reused by both `BaseModel.async_save_hf()` and `BaseComposeModel.async_save_hf()`; each save now only calls  `submit()`.

tests/train/test_trainer_async_hf.py
caixianzhang: Added comments before the pending async HF check in both `BaseModel` and `BaseComposeModel`.

The pending state is used to prevent multiple in-flight async HF saves.Each async HF save stages tensors in CPU memory before the writer process flushes them to disk, so if background I/O is slower than the training loop, allowing multiple pending HF saves would cause staged tensors to accumulate quickly and increase CPU memory pressure.

xtuner/v1/model/base.py
caixianzhang: Agreed. I will keep the current Process-based implementation in this PR to avoid a larger refactor, while keeping the external async HF API as `Future[Path]`. We can evaluate `ProcessPoolExecutor` as a follow-up simplification.

xtuner/v1/model/base.py
caixianzhang: Addressed by adding a configurable timeout to the async HF writer `process.join()`, defaulting to 7200s via
  `ASYNC_HF_WRITER_JOIN_TIMEOUT`; timeout is reported as `child_timeout` and then caught by `AsyncSaveMonitor`.

  For async DCP, the registered future is also watched with a configurable timeout, defaulting to 7200s via `ASYNC_DCP_SAVE_TIMEOUT`; on timeout, the monitor reports a `TimeoutError` and terminates the training process group.

xtuner/v1/model/base.py
caixianzhang: Addressed by removing the extra barriers around async HF finalization. 
The global status `all_gather` already synchronizes writer completion, so only one barrier is kept after rank0 publishes the final HF directory.

xtuner/v1/train/trainer.py
caixianzhang: Addressed by adding `_init_async_save_resources()` and moving async HF/DCP pending state, `AsyncSaveMonitor`, and finalize lock initialization into this internal helper.

xtuner/v1/train/trainer.py
caixianzhang: Addressed by moving async DCP pending wait and checkpoint PG cleanup into the `if self._async_checkpoint:` block, matching the async HF cleanup style.

xtuner/v1/train/trainer.py
caixianzhang: The timeout handling is now in `TrainEngine.async_save_dcp()`.

  The returned future is a commit future, and its commit thread waits on the underlying DCP future with
  `dcp_future.result(timeout=_get_async_dcp_save_timeout())`, defaulting to 7200s via `ASYNC_DCP_SAVE_TIMEOUT`. If it times out, the commit future fails and `AsyncSaveMonitor` catches that failed future and terminates the training process group.

xtuner/v1/train/trainer.py
caixianzhang: Addressed by moving DCP meta update, checkpoint maxkeep cleanup, meta write, and after-save hooks into `_finalize_dcp_save()`.

For async DCP, `_finalize_dcp_save()` is now invoked from the DCP future done callback after `done_future.result()` succeeds, so `meta.json` is only updated after DCP weights have finished saving.

xtuner/v1/utils/async_save_monitor.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def stop(self) -> None:
+        self._stop_event.set()
+        self._monitor_future.result()
+        self._executor.shutdown(wait=True)
+
+    def _run(self) -> None:
+        while True:
+            stopped = self._stop_event.wait(self._interval)
+            if stopped:
```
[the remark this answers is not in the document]
caixianzhang: Verified. The final `AsyncSaveMonitor` implementation defines `_check_watched_futures()`, and `tests/utils/
  test_async_save_monitor.py` calls that same method. There is no `_poll_once` reference left in the final code.

xtuner/v1/utils/process.py
```diff
@@ -26,8 +26,8 @@ def get_async_save_io_priority() -> int | None:
     return _get_optional_int_env("XTUNER_ASYNC_SAVE_IO_PRIORITY", 3)
 
 
-def get_async_save_file_lock_slots() -> int:
-    return max(0, int(os.environ.get("ASYNC_DCP_FILE_WRITE_LOCK_SLOTS", "1")))
+def get_async_hf_save_file_lock_slots() -> int:
+    return max(0, int(os.environ.get("ASYNC_HF_FILE_WRITE_LOCK_SLOTS", "1")))
```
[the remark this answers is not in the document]
caixianzhang: The final env var name is intentional: async HF now uses `ASYNC_HF_FILE_WRITE_LOCK_SLOTS`, aligned with the existing DCP env var `ASYNC_DCP_FILE_WRITE_LOCK_SLOTS`.

 The PR description was stale from an intermediate revision and has been updated to remove the old `XTUNER_...` name and compatibility alias.

tests/utils/test_async_save_monitor.py
```diff
@@ -0,0 +1,48 @@
...
+    def test_removes_successful_future(self):
+        future: Future = Future()
+        future.set_result(Path("/tmp/hf-1"))
+        item = _watch_item(future)
+        monitor = AsyncSaveMonitor()
+
+        monitor.register(item)
+        with patch("xtuner.v1.utils.async_save_monitor.os.killpg") as killpg:
+            monitor._check_watched_futures()
+
```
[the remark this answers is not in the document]
caixianzhang: Verified against the current final diff. `AsyncSaveMonitor` defines `_check_watched_futures()`, and `tests/utils/
  test_async_save_monitor.py` calls that same method. There is no `_poll_once()` reference left in the current code, so the test method name is consistent.

tests/utils/test_async_save_monitor.py
```diff
@@ -0,0 +1,48 @@
...
+    def test_terminates_process_group_on_future_failure(self):
+        future: Future = Future()
+        future.set_exception(RuntimeError("mock async save failure"))
+        item = _watch_item(future)
+        monitor = AsyncSaveMonitor()
+
+        monitor.register(item)
+        with (
+            patch("xtuner.v1.utils.async_save_monitor.os.getpgrp", return_value=1234) as getpgrp,
```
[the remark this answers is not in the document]
caixianzhang: Verified. The current `AsyncSaveMonitor` implementation defines `_check_watched_futures()`, and this test calls the same method. `_poll_once()` is not present in the current final code, so no test update is needed.

tests/utils/test_async_save_monitor.py
```diff
@@ -0,0 +1,48 @@
...
+            patch("xtuner.v1.utils.async_save_monitor.os.killpg") as killpg,
+        ):
+            monitor._check_watched_futures()
+
+        getpgrp.assert_called_once_with()
+        killpg.assert_called_once_with(1234, signal.SIGTERM)
```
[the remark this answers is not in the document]
caixianzhang: Thanks for checking. Confirmed, the current implementation and tests  both use `_check_watched_futures()`.

xtuner/v1/patch/xtuner_storage.py
```diff
@@ -70,7 +70,7 @@ def _release_write_lock(fd: int) -> None:
 
 
 def _get_file_write_lock_slots() -> int:
-    return max(0, int(os.environ.get("ASYNC_DCP_FILE_WRITE_LOCK_SLOTS", "0")))
+    return max(0, int(os.environ.get("ASYNC_DCP_FILE_WRITE_LOCK_SLOTS", "1")))
```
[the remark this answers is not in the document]
caixianzhang: This change is intentional based on validation feedback. Async DCP background writes can contend with async HF and training for shared storage / CPU I/O, so the default `ASYNC_DCP_FILE_WRITE_LOCK_SLOTS` is changed to `1` to throttle DCP file writes.

I updated the PR description to explicitly mention this DCP default behavior change.

HAOCHENYE (approved):

PR #1878, opened 2026-06-05
[Fix] DCP async checkpoint daemon TCPStore fails with EADDRINUSE during spawn window

VincentCheungKokomo:
[Fix] Pin DCP async checkpoint daemon port outside the ephemeral range

  The process-based async checkpoint picks the daemon TCPStore port on rank0
  via get_free_port (bind(0) -> ephemeral port -> close), then binds it later
  in a spawned subprocess. In that spawn window a transient connection's
  ephemeral source port can grab the freed port, so the daemon listen fails
  with EADDRINUSE -- more likely at scale, where the daemon group's own
  bring-up bursts ephemeral connects on node0.
  
  Route the daemon port to a fixed range outside the kernel ephemeral range
  (default [29600, 30000)); the kernel never auto-assigns source ports there,
  so it can't be grabbed regardless of scale. Drop the now-unnecessary,
  cross-rank deadlock-prone EADDRINUSE retry in favor of fail-fast.

### DIFF
--- a/tests/utils/test_dcp_async_port.py
+++ b/tests/utils/test_dcp_async_port.py
@@ -1,0 +1,139 @@
+"""Regression tests for the DCP async-checkpoint daemon port patch.
+
+The original bug: PyTorch's process-based async checkpoint picks the daemon TCPStore
+port via ``get_free_port`` (``bind(("localhost", 0))`` -> ephemeral port -> close),
+so a transient outbound connection can grab the same port during the spawn window and
+the daemon ``listen`` fails with ``EADDRINUSE`` (more likely at large scale).
+
+``patch_dcp_async_daemon_port`` relocates the daemon port to a fixed range *outside*
+the kernel ephemeral range, which the kernel never auto-assigns. These tests verify the
+patch is wired to the symbol torch actually calls and that the chosen port is kept out
+of the ephemeral range.
+"""
+
+import socket
+from unittest.mock import MagicMock
+
+import pytest
+import torch.distributed.checkpoint._async_process_executor as _async_process_executor
+
+import xtuner.v1.patch.dcp_async_port as dcp_async_port
+from xtuner.v1.patch import patch_dcp_async_daemon_port
+
+
+def _free_port_for_test() -> int:
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        sock.bind(("", 0))
+        return sock.getsockname()[1]
+    finally:
+        sock.close()
+
+
+@pytest.fixture
+def reset_patch_state():
+    original = _async_process_executor.get_free_port
+    was_patched = dcp_async_port._PATCHED
+    dcp_async_port._PATCHED = False
+    yield
+    _async_process_executor.get_free_port = original
+    dcp_async_port._PATCHED = was_patched
+
+
+class TestDcpAsyncPort:
+    def test_patch_replaces_get_free_port(self, reset_patch_state):
+        assert _async_process_executor.get_free_port is not dcp_async_port._xtuner_dcp_get_free_port
+        patch_dcp_async_daemon_port()
+        # torch calls the name bound in the executor module namespace, so that is what
+        # must be replaced for the fix to take effect.
+        assert _async_process_executor.get_free_port is dcp_async_port._xtuner_dcp_get_free_port
+
+    def test_patch_is_idempotent(self, reset_patch_state):
+        patch_dcp_async_daemon_port()
+        sentinel = lambda: 0  # noqa: E731 - stand-in for a later reassignment
+        _async_process_executor.get_free_port = sentinel
+        patch_dcp_async_daemon_port()  # second call must be a no-op
+        assert _async_process_executor.get_free_port is sentinel
+
+    def test_chosen_port_is_outside_ephemeral(self, monkeypatch):
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_BASE", "29600")
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_SPAN", "400")
+        monkeypatch.setattr(dcp_async_port, "_ephemeral_range", lambda: (32768, 60999))
+        port = dcp_async_port._xtuner_dcp_get_free_port()
+        assert 29600 <= port < 30000
+        assert port < 32768
+
+    def test_env_overrides_range(self, monkeypatch):
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_BASE", "31234")
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_SPAN", "7")
+        assert dcp_async_port._port_base() == 31234
+        assert dcp_async_port._port_span() == 7
+
+    def test_port_span_clamped_to_minimum_one(self, monkeypatch):
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_SPAN", "0")
+        assert dcp_async_port._port_span() == 1
+
+    def test_probe_skips_occupied_port(self, monkeypatch):
+        occupied = _free_port_for_test()
+        listener = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        listener.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        listener.bind(("", occupied))
+        listener.listen(1)
+        try:
+            monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_BASE", str(occupied))
+            monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_SPAN", "5")
+            monkeypatch.setattr(dcp_async_port, "_ephemeral_range", lambda: None)
+            port = dcp_async_port._xtuner_dcp_get_free_port()
+            # SO_REUSEADDR still cannot bind over an actively listening socket, so the
+            # probe must move past the occupied base port.
+            assert occupied < port < occupied + 5
+        finally:
+            listener.close()
+
+    def test_no_free_port_raises(self, monkeypatch):
+        occupied = _free_port_for_test()
+        listener = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        listener.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        listener.bind(("", occupied))
+        listener.listen(1)
+        try:
+            monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_BASE", str(occupied))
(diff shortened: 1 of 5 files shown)


tests/utils/test_dcp_async_port.py
```diff
@@ -0,0 +1,139 @@
...
+        patch_dcp_async_daemon_port()  # second call must be a no-op
+        assert _async_process_executor.get_free_port is sentinel
+
+    def test_chosen_port_is_outside_ephemeral(self, monkeypatch):
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_BASE", "29600")
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_SPAN", "400")
+        monkeypatch.setattr(dcp_async_port, "_ephemeral_range", lambda: (32768, 60999))
+        port = dcp_async_port._xtuner_dcp_get_free_port()
+        assert 29600 <= port < 30000
+        assert port < 32768
```
HAOCHENYE: Tests should be code-agnostic and should avoid testing internal properties as much as possible. Tests should focus on functionality rather than the code implementation itself.

tests/utils/test_dcp_async_port.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def test_env_overrides_range(self, monkeypatch):
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_BASE", "31234")
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_SPAN", "7")
+        assert dcp_async_port._port_base() == 31234
```
HAOCHENYE: Same as above

tests/utils/test_dcp_async_port.py
```diff
@@ -0,0 +1,139 @@
...
+        assert dcp_async_port._port_span() == 7
+
+    def test_port_span_clamped_to_minimum_one(self, monkeypatch):
```
HAOCHENYE: Same as above

tests/utils/test_dcp_async_port.py
```diff
@@ -0,0 +1,139 @@
...
+        monkeypatch.setenv("XTUNER_DCP_DAEMON_PORT_SPAN", "0")
+        assert dcp_async_port._port_span() == 1
+
+    def test_probe_skips_occupied_port(self, monkeypatch):
```
HAOCHENYE: Same as above

tests/utils/test_dcp_async_port.py
```diff
@@ -0,0 +1,139 @@
...
+        "base,hi_excl,eph,expected",
+        [
+            (29600, 30000, (32768, 60999), True),  # fully below ephemeral range
+            (61000, 61400, (32768, 60999), True),  # fully above ephemeral range
+            (40000, 40400, (32768, 60999), False),  # inside ephemeral range
+            (29600, 30000, None, False),  # unknown range -> conservative False
+            (32768, 33000, (32768, 60999), False),  # overlaps the lower bound
+        ],
+    )
+    def test_is_outside_ephemeral(self, base, hi_excl, eph, expected):
```
HAOCHENYE: The variable names here are excessively abbreviated

tests/utils/test_dcp_async_port.py
HAOCHENYE, line 1: Please test the functionality of public interfaces, and avoid testing against internal unstable interfaces as much as possible

xtuner/v1/patch/dcp_async_port.py
```diff
@@ -0,0 +1,149 @@
...
+        return None
+
+
+def _is_outside_ephemeral(base: int, hi_excl: int, eph: tuple[int, int] | None) -> bool:
+    """Whether [base, hi_excl) does not intersect the kernel ephemeral range.
+
+    The kernel only auto-assigns outbound ports from the ephemeral range, so a daemon port outside it cannot be grabbed
+    by a transient connection during the spawn window.
+    """
+    if eph is None:
```
HAOCHENYE: Avoid excessive abbreviation

xtuner/v1/patch/dcp_async_port.py
```diff
@@ -0,0 +1,149 @@
...
+    (net.ipv4.ip_local_port_range)."""
+    try:
+        with open("/proc/sys/net/ipv4/ip_local_port_range") as f:
+            lo, hi = f.read().split()[:2]
+            return int(lo), int(hi)
+    except Exception:
+        return None
+
+
+def _is_outside_ephemeral(base: int, hi_excl: int, eph: tuple[int, int] | None) -> bool:
```
HAOCHENYE: Avoid excessive abbreviation

xtuner/v1/patch/dcp_async_port.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    base = _port_base()
+    span = _port_span()
+    eph = _ephemeral_range()
+    last_err: OSError | None = None
+    for port in range(base, base + span):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            sock.bind(("", port))
```
HAOCHENYE, line 119: https://github.com/pytorch/pytorch/blob/4fc800d2744268635e4b5b96b81de6955219410c/torch/distributed/elastic/rendezvous/etcd_server.py#L58

Why not bind localhost here? Is there any reason?

xtuner/v1/patch/dcp_async_port.py
```diff
@@ -0,0 +1,149 @@
...
+            sock.listen(1)
+            # Per-checkpoint evidence for verification: every line must say "outside"; any
+            # "in-ephemeral" means the race can still trigger. Greppable at scale.
+            status = "outside" if _is_outside_ephemeral(port, port + 1, eph) else "in-ephemeral"
+            logger.info(f"[DCP async_save] DCP daemon port chosen: {port} ({status} ephemeral range {eph})")
```
HAOCHENYE: Use `logger.debug`

xtuner/v1/patch/dcp_async_port.py
```diff
@@ -0,0 +1,149 @@
...
+# Default daemon port range, outside the ephemeral range (>=32768) and clear of the
+# common training MASTER_PORT (29500).
+_DEFAULT_PORT_BASE = 29600
+_DEFAULT_PORT_SPAN = 400
+
+_PATCHED = False
+
+
+def _port_base() -> int:
+    return int(os.environ.get("XTUNER_DCP_DAEMON_PORT_BASE", str(_DEFAULT_PORT_BASE)))
```
HAOCHENYE: Don't set two environment variables; it's too complicated. Just use one environment variable, and set it with something like xxxx="26000:30000"

xtuner/v1/patch/dcp_async_port.py
```diff
@@ -0,0 +1,149 @@
...
+    if _PATCHED:
+        return
+    _async_process_executor.get_free_port = _xtuner_dcp_get_free_port
+    _PATCHED = True
+
+    base = _port_base()
+    hi_excl = base + _port_span()
+    eph = _ephemeral_range()
+    if _is_outside_ephemeral(base, hi_excl, eph):
+        logger.info(
```
HAOCHENYE: logger.debug

xtuner/v1/patch/dcp_async_port.py
```diff
@@ -0,0 +1,149 @@
...
+    base = _port_base()
+    span = _port_span()
+    eph = _ephemeral_range()
+    last_err: OSError | None = None
+    for port in range(base, base + span):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            sock.bind(("", port))
+            sock.listen(1)
```
HAOCHENYE: Similarly, why is listen(1) used here instead of listen(0)?

xtuner/v1/patch/dcp_async_port.py
VincentCheungKokomo: The daemon subprocess inits a GLOO PG with MASTER_ADDR=node0's FQDN (not localhost), so its TCPStore server binds all interfaces to be reachable cross-node. Probing the wildcard ("", port) is therefore the faithful predictor of the real daemon bind — probing localhost would miss occupancy on the external interface. etcd_server binds localhost because it's a single-host rendezvous; this daemon is cross-node. Added a comment to make this explicit.

xtuner/v1/patch/dcp_async_port.py
VincentCheungKokomo: Done — replaced the two vars with a single XTUNER_DCP_DAEMON_PORT_RANGE="29600:30000" (high exclusive); default range is unchanged. Updated parser, error message, and tests.

xtuner/v1/patch/dcp_async_port.py
VincentCheungKokomo: The probe socket is closed immediately, so the backlog value is irrelevant — we only need to confirm the port can enter LISTEN. Changed to listen(0) and added a clarifying comment.

tests/utils/test_dcp_async_port.py
VincentCheungKokomo: removed

tests/utils/test_dcp_async_port.py
VincentCheungKokomo: removed

tests/utils/test_dcp_async_port.py
VincentCheungKokomo: removed

tests/utils/test_dcp_async_port.py
VincentCheungKokomo: removed

xtuner/v1/patch/dcp_async_port.py
VincentCheungKokomo: changed to logger.debug

xtuner/v1/patch/dcp_async_port.py
VincentCheungKokomo: changed to logger.debug

xtuner/v1/patch/dcp_async_port.py
VincentCheungKokomo: variable renamed

xtuner/v1/patch/dcp_async_port.py
VincentCheungKokomo: variable renamed

tests/utils/test_dcp_async_port.py
VincentCheungKokomo: test removed

HAOCHENYE:
@claude review

HAOCHENYE (approved):

PR #1880, opened 2026-06-05
[Enhance] avoid peak memory in internal metrics dummy fwd

nil0x9:
Pass `loss_ctx` through the internal metrics dummy forward path.

For long-sequence training, passing `loss_ctx=None` into model forward would materialize a large fp32 `logits` tensor.
https://github.com/InternLM/xtuner/blob/0e40b0d7ba11b7357d3fcff05c24f171ad8bf972/xtuner/v1/module/lm_head/lm_head.py#L46-L47

This will lead a large peak memory spike and even OOM. Passing in a dummy `loss_ctx` keeps memory footprint within the range of a regular forward pass (typically chunked loss calculation).


nil0x9:
@claude review

HAOCHENYE (approved):

PR #1881, opened 2026-06-05
[Enhance] Muon: AGRS uneven sharding, peak mem reduction, refactor async tasks and add all2all flag

nil0x9:
## Summary

  - Refactor `muon_update_batch_async` to reduce peak memory usage and consolidate process-group arguments
  - Extend AGRS to support uneven sharding — previously, remainder batches with uneven shard sizes fell back to
  all-to-all; AGRS now handles them internally via padding, keeping all remainder batches on the AGRS path
  - Defer pad_batch allocation into the async task body via a new batch_size parameter
  - Add `enable_all2all=True` flag to Muon and MuonConfig; when disabled, full batches that would otherwise use all-to-all fall back to AGRS, for cluster topologies where all-to-all is unreliable
  - Replace `test_muon_compile.py` with a new end-to-end test covering single-GPU and 4-GPU FSDP cases, validating all four communication strategies against a reference implementation


nil0x9:
@claude review

nil0x9:
@claude previous review comments are addressed. review again.

HAOCHENYE (approved):

PR #1907, opened 2026-06-11
[dev] add offline script to collect internal metrics

nil0x9:


nil0x9:
@claude it is a utility script that does not affect main feature or ci of the repo. merge it directly

PR #1937, opened 2026-06-29
Dsv4

HAOCHENYE:


docs/design/mtp_loss_context.md
HAOCHENYE, line 1: drop it, unnecessary changes.

docs/design/mtp_sp_support.md
HAOCHENYE, line 1: drop it, unnecessary changes.

tests/model/test_deepseek_v4_moe.py
```diff
@@ -0,0 +1,260 @@
...
+            if not unmatched:
+                covered += 1
+            else:
+                missing.append((xname, unmatched))
+
+        # Allow ad-hoc gaps for MTP-side translation (which is best-effort in
+        # PR9; the model's `mtp_config=None` here means none should appear) but
+        # require ≥ 90% coverage for the main stack.
+        coverage_ratio = covered / len(xtuner_names)
+        assert coverage_ratio >= 0.90, (
```
HAOCHENYE: If it is only to skip the mtp check, please explicitly ensure that all parameters before mtp are fully loaded, rather than using magic numbers like 90%, and indicate that the mtp test is a TODO item

tests/model/test_deepseek_v4_moe.py
```diff
@@ -0,0 +1,260 @@
...
+        assert cfg.attention.num_key_value_heads == 1
+        # HC config.
+        assert cfg.hc_cfg.hc_mult == 4
+        assert cfg.hc_cfg.hc_sinkhorn_iters == 20
+        assert cfg.swiglu_limit == 10.0
+        # MTP wired from `num_nextn_predict_layers`.
+        assert cfg.mtp_config is not None
+        assert cfg.mtp_config.num_layers == 1
+
+    def test_to_hf_key_list_coverage(self) -> None:
```
HAOCHENYE: The testing method for parameter alignment is too indirect. It is recommended to adopt the following testing approach:

Load a DeepSeekv4-flash model with only one layer （num_hidden_layers=1）, then execute load and save, and test whether the keys obtained from load and save align with the original weights at the bitwisze level

tests/model/test_deepseek_v4_moe.py
HAOCHENYE, line 1: Combine this test and the layer_parity test into one file

tests/model/test_deepseek_v4_decoder_layer_parity.py
```diff
@@ -0,0 +1,1178 @@
...
+    # over bf16 routing; we only pin to ``"native"`` here to make parity
+    # measurable.)
+    cfg.router_compute_dtype = "native"
+    return cfg
+
+
+# ─── HF → XTuner weight copy ────────────────────────────────────────────────────
+
+
+def _copy_hf_to_xtuner_layer(
```
HAOCHENYE: Why is it so complicated? Can we make toy model save_pretrained and then let xtuner load from it? This way we wouldn't need to hardcode like this

tests/model/test_deepseek_v4_decoder_layer_parity.py
```diff
@@ -0,0 +1,1178 @@
...
+            "heavily_compressed_attention", num_hash_layers=0, layer_idx=1
+        )
+        hf_layer = hf_model.layers[layer_idx]
+        _install_hf_attention_fallback(xtuner_layer, hf_layer, hf_model)
+        _install_hf_moe_fallback(xtuner_layer, hf_layer)
+        hidden_states, input_ids, position_ids = self._common_inputs(seq_len=256)
+        with torch.no_grad():
+            hf_out = _run_hf_layer(hf_model, layer_idx, hidden_states, input_ids, position_ids)
+            xt_out = _run_xtuner_layer(xtuner_layer, hidden_states, input_ids, position_ids, hf_model)
+        torch.testing.assert_close(xt_out, hf_out, atol=atol, rtol=atol)
```
HAOCHENYE: Here are a few DeepSeekv4 tests:

1. Run hash router layer alone
2. Run csa alone
3. Run hca alone
4. Run a minimum DeepSeekv4-flash, these layers all have

Finally, all tests should go through the HF path to ensure bitwise alignment!

tests/model/test_deepseek_v4_decoder_layer_parity.py
```diff
@@ -0,0 +1,1178 @@
...
+
+    def test_subcomponent_probe(self, capsys) -> None:
```
HAOCHENYE: I completely don't get the point of this test? If it's just for temporary debugging to show where it forked, please comment it out

tests/module/dsa/test_dsa.py
HAOCHENYE, line 1: All op and submodule level tests have the only reference being the Hugging Face implementation, and then control the absolute and relative errors

xtuner/v1/utils/compile.py
```diff
@@ -15,6 +16,121 @@
 TARGET_DEVICE = get_device()
 
 
+def _patch_sympy_mod_eval_negative_subs() -> None:
```
HAOCHENYE, line 19: Double check is this necessary

xtuner/v1/utils/compile.py
```diff
@@ -15,6 +16,121 @@
...
+        return None
+
+    _safe_eval.__xtuner_negative_subs_patch__ = True  # type: ignore[attr-defined]
+    setattr(Mod, "eval", classmethod(_safe_eval))
+
+
+_patch_sympy_mod_eval_negative_subs()
+
+
+def _patch_triton_max_block_xblock() -> None:
```
HAOCHENYE, line 92: Same as above

PR #1959, opened 2026-07-16
[Draft] core feature refine

HAOCHENYE:

PR #1961, opened 2026-07-17
[FSDP] Add reduce-sum gradient reduction helper

HAOCHENYE:
**This PR (1/5)** — behavior-neutral. Adds `BaseModel.set_gradient_reduce_sum()`, which pairs `set_gradient_divide_factor(1.0)` with `set_force_sum_reduction_for_comms(True)` over every `FSDPModule` (torch>=2.10 guarded, avoids the bf16 NCCL PreMulSum zeroing trap). Not wired into any `fully_shard` yet.

Full stack (merge bottom-up, under #1959):
1. `reduce-sum-01-fsdp-helper` — [FSDP] Add reduce-sum gradient reduction helper
2. `reduce-sum-02-split-logging` — [Loss] Split logging loss from backward loss
3. `reduce-sum-03-switch-sum` — [FSDP][Loss] Switch gradient reduction to SUM
4. `reduce-sum-04-remove-world-size` — [Refactor] Remove unused world_size plumbing
5. `reduce-sum-05-drop-nonglobal` — [Refactor] Drop non-global aux-loss averaging mode

### DIFF
--- a/tests/model/test_reduce_sum_grad.py
+++ b/tests/model/test_reduce_sum_grad.py
@@ -1,0 +1,76 @@
+import torch
+import torch.distributed as dist
+import torch.nn as nn
+from torch.distributed.fsdp import MixedPrecisionPolicy, fully_shard
+
+from xtuner._testing.testcase import DeterministicDDPTestCase
+from xtuner.v1.model.base import BaseModel, XTunerBaseModelConfig
+
+
+class _ReduceSumToyConfig(XTunerBaseModelConfig):
+    hidden_size: int = 8
+
+    def build(self) -> "_ReduceSumToyModel":
+        return _ReduceSumToyModel(self)
+
+
+class _ReduceSumToyModel(BaseModel):
+    config: _ReduceSumToyConfig
+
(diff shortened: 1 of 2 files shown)


tests/model/test_reduce_sum_grad.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class _ReduceSumToyConfig(XTunerBaseModelConfig):
+    hidden_size: int = 8
+
+    def build(self) -> "_ReduceSumToyModel":
+        return _ReduceSumToyModel(self)
+
+
+class _ReduceSumToyModel(BaseModel):
```
HAOCHENYE, line 17: This test is too simplistic; it needs to have parameters with both replicate and shard forms. However, since the model does not yet implement the corresponding gradient calibration code (shard is not processed, replicate does reduce sum?), it is recommended to only add the test in PR 1963 and not implement it here.

HAOCHENYE:
Superseded by the consolidated 2-PR stack #1968 (core) + #1969 (drop non-global). Closing this fragmented stack.

PR #1965, opened 2026-07-17
[Refactor] Drop non-global aux-loss averaging mode

HAOCHENYE:
**This PR (5/5)** — breaking config change. Drops the `balancing_loss_global_average` / `z_loss_global_average` switches and the non-global aux-loss mode, which is incompatible with reduce-sum (SUM-aggregated replicated router grad would inflate by world_size). Distributed = global-average; single-process is its W=1 special case.

Full stack (merge bottom-up, under #1959):
1. `reduce-sum-01-fsdp-helper` — [FSDP] Add reduce-sum gradient reduction helper
2. `reduce-sum-02-split-logging` — [Loss] Split logging loss from backward loss
3. `reduce-sum-03-switch-sum` — [FSDP][Loss] Switch gradient reduction to SUM
4. `reduce-sum-04-remove-world-size` — [Refactor] Remove unused world_size plumbing
5. `reduce-sum-05-drop-nonglobal` — [Refactor] Drop non-global aux-loss averaging mode

### DIFF
--- a/docs/design/reduce_sum_grad.md
+++ b/docs/design/reduce_sum_grad.md
@@ -1,0 +1,176 @@
+# FSDP 梯度 reduce-sum 改造设计
+
+## 1. 背景
+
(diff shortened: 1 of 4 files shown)


HAOCHENYE:
Superseded by the consolidated 2-PR stack #1968 (core) + #1969 (drop non-global). Closing this fragmented stack.

PR #1968, opened 2026-07-20
[FSDP][Loss] Reduce-sum gradient reduction with per-batch aux-loss hub

HAOCHENYE:
Switch FSDP gradient reduction from mean to pure **SUM** and remove the compensating ×world_size injections across CE / balancing / z-loss, so each parameter's gradient equals the global loss gradient with no compensating divides. Squashed to a single commit (the intermediate steps had incomplete calibration and would not individually pass CI).

- **FSDP reduce-scatter → SUM**: `set_gradient_divide_factor(1.0)` + `set_force_sum_reduction_for_comms(True)` (avoids the bf16 NCCL PreMulSum zeroing), wired into every `fully_shard` incl. compose/InternS1; `scale_and_reduce_grad` drops the expert `div_(ep_size)` / replicated pre-divide and SUM-reduces `Replicate`-placement (incl. fp32-ignored) params.
- **Loss**: drop the CE WORLD `all_reduce`, balancing `all_reduce_autograd`, z-loss `× world_size`, and the dead `world_size` plumbing.
- **Display**: per-rank `calibrate()` with no cross-rank all_reduce — each rank shows its own per-token-mean loss.
- **Aux losses** folded into a per-batch `AuxLossContext` hub (`MoELossContextDict = {lm, aux, mtp}`); `AuxLossConfig.build(data, sp_mesh)` conforms to the loss-config contract; per-layer `accumulate` takes router-output lists and does padding removal / histc / balancing / z internally.
- **Drop** the non-global aux-loss averaging mode (incompatible with reduce-sum).

Tests: `tests/model/test_reduce_sum_grad.py` (16) — bf16 SUM mechanism, EP1/EP2 token-mean parity, balancing & z-loss A/B, compose hook, Dense fp32 reduction, domino-MTP, SP world=4, per-rank display. pre-commit clean.

Supersedes #1969 (folded in).

### DIFF
--- a/xtuner/v1/engine/train_engine.py
+++ b/xtuner/v1/engine/train_engine.py
@@ -35,7 +35,6 @@ from xtuner.v1.model.base import (
     BatchForwardInfo,
     DataBatchInfo,
     ModelItem,
-    ModelOutputs,
     XTunerBaseModelConfig,
 )
 from xtuner.v1.patch.xtuner_storage import XtunerCacheWriter, _get_async_dcp_save_timeout
@@ -51,7 +50,7 @@ from xtuner.v1.utils.grad_norm import cal_grad_norm
 
 
 class TrainStepInfo(DataBatchInfo, BatchForwardInfo):
-    total_loss: float
+    pass
 
 
 logger = get_logger()
@@ -216,7 +215,6 @@ class TrainEngine:
         micro_batch_results = []
 
         data_batch_info = self.model.pre_micro_batch_forward(data_batches)
-        total_loss = torch.tensor(0.0, device=DEVICE)
 
         for i in range(0, len(data_batches), intra_layer_micro_batch):
             ProberList.set_micro_batch_iter(micro_batch_iter)
@@ -238,14 +236,15 @@ class TrainEngine:
 
             micro_batch_results.append(output)
 
-            loss = self._get_total_loss(output)
-            loss.backward()
-            total_loss += loss.detach()
+            # The model folds every loss term (LM/policy + aux + MTP) into output.loss; the engine
+            # just backprops it. Per-term display values ride on output.loss_log.
+            assert output.loss is not None, "model forward must return a loss in training mode"
+            output.loss.backward()
             # call dump_forward_records after backward to record the recomputed activations
             ProberList.after_micro_iter_forward()
 
         batch_forward_info = self.model.post_micro_batch_forward(micro_batch_results)
-        return TrainStepInfo(total_loss=total_loss.item(), **data_batch_info, **batch_forward_info)
+        return TrainStepInfo(**data_batch_info, **batch_forward_info)
 
     def from_hf(self, hf_path: str | Path, strict: bool = False):
         self.model.from_hf(hf_path=hf_path, strict=strict)
@@ -558,17 +557,3 @@ class TrainEngine:
         for model in self.model.modules():
             if isinstance(model, BaseModel) and model.float8_handler is not None:
                 model.float8_handler.precompute_float8_dynamic_scale_for_fsdp(model)
-
-    def _get_total_loss(self, model_outputs: ModelOutputs) -> torch.Tensor:
-        # TODO: This logic should be moved into the model layer. The model should be responsible
-        # for aggregating all losses (CE loss, balancing loss, z loss, etc.) and returning a
-        # single total_loss. The engine should only call model.forward() and use the returned
-        # total_loss directly, rather than iterating through fields to sum losses here.
-        # This would provide better separation of concerns and make the loss computation logic
-        # more explicit and maintainable.
-        loss = torch.tensor(0.0, device=DEVICE)
-        for key in model_outputs.model_fields:
-            value = getattr(model_outputs, key)
-            if "loss" in key and isinstance(value, torch.Tensor):
-                loss += value
-        return loss
--- a/xtuner/v1/loss/ce_loss.py
+++ b/xtuner/v1/loss/ce_loss.py
@@ -6,7 +6,6 @@ import torch.distributed as dist
 import torch.nn.functional as F
 from cyclopts import Parameter
 from torch.distributed.device_mesh import DeviceMesh
-from torch.distributed.nn.functional import all_reduce
 
 from xtuner.v1.utils.device import get_device
 
@@ -106,6 +105,11 @@ class LMHeadLossContext(BaseLossContext):
     loss_cfg: CELossConfig
     loss_kwargs: CELossKwargs
 
+    # Display state for calibrate(): coefficient stashed by build_batches (global_denom /
+    # rank_denom) and this rank's detached backward loss stashed by forward().
+    _display_coeff: torch.Tensor
+    _backward_loss: torch.Tensor
+
     def __init__(self, loss_cfg: CELossConfig, loss_kwargs: CELossKwargs):
         super().__init__(loss_cfg, loss_kwargs)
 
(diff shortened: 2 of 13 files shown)


xtuner/v1/loss/ce_loss.py
```diff
@@ -178,10 +182,21 @@ def build_batches(  # type: ignore[override]
...
 
+        # Coefficient to turn each rank's local backward loss (local token sum / global denominator)
+        # back into this rank's per-token mean for display: multiply by global_denominator, divide by
+        # this rank's own grad-token count across the whole step. Summing `calibrate()` over the
+        # micro-batches then yields this rank's per-token-mean loss with no cross-rank all_reduce, and
+        # equals the global loss when world size is 1 (global_denominator == step_grad_tokens).
+        step_grad_tokens = torch.zeros((), device=global_denominator.device)
+        for loss_ctx in loss_ctx_list:
+            step_grad_tokens += (loss_ctx.loss_kwargs.shifted_labels != loss_cfg.ignore_idx).sum()
+        display_coeff = (global_denominator / step_grad_tokens.clamp_min(1.0)).detach()
```
HAOCHENYE: Shouldn't the coefficient here just be the global dominator divided by the rank dominator after the reduce sum? I think there is an issue with your current algorithm; calibrating at the token level this way is fine, but it doesn't work for sample-level and square-level losses.

xtuner/v1/model/moe/moe.py
```diff
@@ -828,29 +879,33 @@ def _forward(
...
             z_ctx=z_ctx,
             non_pad_token=non_pad_token,
         )
-        balancing_loss, z_loss, tokens_per_expert_global = split_aux_output
-        if balancing_loss is not None:
-            output["balancing_loss"] = balancing_loss
-        if z_loss is not None:
-            output["z_loss"] = z_loss
-        output["tokens_per_expert_global"] = tokens_per_expert_global
+        if aux_out["balancing_loss"] is not None:
```
HAOCHENYE: I think we can update the contract for the model's output. The `loss` field in the output should represent the sum of all losses, with the engine layer handling simple gradient accumulation and the backward pass. Then, we can add a new `loss_log` field to store all the calibrated results, which the engine layer can simply sum up and include in the logs. Does that sound reasonable to you?

Also, there is currently a lot of hardcoding in `post_micro_batch_forward`, and the field names aren't very logical. My vision for the final format is: recording both local loss and reduced loss in TensorBoard, with logs appearing as `loss(reduced): 1.2(1.3)`.

xtuner/v1/model/base.py
```diff
@@ -576,7 +580,31 @@ def from_hf(
...
-        return
+        # Params excluded from FSDP sharding (e.g. fp32 ``ignored_params`` matched by
+        # ``fp32_keys_pattern``) are distributed as ``Replicate`` DTensors, so they get no
+        # reduce-scatter. Under reduce-sum their gradient is this rank's local component; without a
+        # cross-rank reduction the replicated copies would carry different grads and diverge. SUM
+        # (no divide) their grads over the replicate group here. FSDP-sharded params are handled by
+        # the reduce-scatter and are skipped (no Replicate placement).
+        self._reduce_replicated_ignored_grads(self.trainable_parameters())
+
+    def _reduce_replicated_ignored_grads(self, params: Iterable[tuple[str, nn.Parameter]]) -> None:
```
HAOCHENYE, line 593: I'm a bit confused here. Could you consider adding a comment? If I have TPSP enabled, an `allgather` is executed before the attention mechanism. If the attention structure includes QK normalization, that norm will process the full dataset. Wouldn't performing an `allreduce` sum on the QK norm gradients cause them to be scaled up by a factor of SP?

xtuner/v1/model/base.py
```diff
@@ -1305,41 +1372,33 @@ def pre_micro_batch_forward(self, data_batches: Sequence[ModelItem]) -> DataBatc
...
-        return ret
+        Returns:
+            dict[str, float]: ``reduced_<name>`` -> this rank's display loss for each term.
+        """
+        summed: dict[str, torch.Tensor] = {}
+        for output in batch_outputs:
+            for name, value in (output.calibrated_losses or {}).items():
+                contribution = value.detach().float()
+                summed[name] = contribution if name not in summed else summed[name] + contribution
+        return {f"reduced_{name}": value.item() for name, value in summed.items()}
```
HAOCHENYE, line 1379: Where does the reduction manifest? My understanding is that this is actually the local loss.

xtuner/v1/engine/train_engine.py
```diff
@@ -240,12 +243,13 @@ def train_step(self, data_batches: list[ModelItem]) -> TrainStepInfo:
 
             loss = self._get_total_loss(output)
             loss.backward()
-            total_loss += loss.detach()
+            for value in (output.calibrated_losses or {}).values():
+                display_total_loss += value.detach().float()
```
HAOCHENYE: Can these operations be moved into `post_micro_batch_forward`? Do you think that’s appropriate?

xtuner/v1/loss/moe_loss.py
HAOCHENYE, line 49: Could we compute `non_pad_tokens` inside `AuxLossConfig.build`? That way, the `accumulate` interface could be further simplified.

xtuner/v1/loss/moe_loss.py
```diff
@@ -139,34 +122,61 @@ def finalize(
...
+            loss_vec = scale_global * (tokens_per_expert_local * routing_weights_mean_global).sum(-1)
+
+        loss = loss_vec.sum() * alpha / self._batch_size
+        # Display value (detached, no all_reduce): this rank's balancing loss from LOCAL statistics
+        # (its own tokens_per_expert / seqlen), a readable per-rank number. This is a display-only
+        # computation, NOT a backward mode -- it does not reintroduce the removed non-global averaging.
+        self._calibrated = self._local_balancing_loss(
+            local_gating_sum, tokens_per_expert_local, n_routed_experts, num_experts_per_tok, non_pad_token
+        )
+        return loss
```
HAOCHENYE, line 197: Is there no way to log the global balancing loss anymore? It seems like it's no longer possible to reflect the global balancing status.

xtuner/v1/model/moe/moe.py
```diff
@@ -227,24 +269,23 @@ def _z_loss_dist_token_count(
...
         first = z_ctx[0] if isinstance(z_ctx, list) else z_ctx
         if not first.loss_cfg.z_loss_global_average or not dist.is_initialized():
-            return None, 1
+            return None
         n = torch.tensor(num_tokens_local, device=device, dtype=torch.int64)
         group = dist.group.WORLD
         assert group is not None
-        n_global = all_reduce(n, "sum", group)
-        return n_global, dist.get_world_size()
+        return all_reduce(n, "sum", group)
```
HAOCHENYE: If `AuxLossConfig` receives `data_batch` during the build phase and constructs the corresponding `loss_kwargs` (e.g., `non_pad_indices`, `local_global_tokens`, etc.), is it still necessary to implement this method within the model?

xtuner/v1/model/moe/moe.py
```diff
@@ -96,6 +96,48 @@
...
+    total = ctxs[0].calibrate()
+    for c in ctxs[1:]:
+        total = total + c.calibrate()
+    return total
+
+
+def _build_calibrated_losses(
+    lm_display: torch.Tensor,
+    balancing_ctx: object | list | None,
+    z_ctx: object | list | None,
```
HAOCHENYE: Why is the type hint `object` instead of the corresponding `LossContext` class?

xtuner/v1/model/moe/moe.py
```diff
@@ -96,6 +96,48 @@
...
+    ctxs = ctx if isinstance(ctx, list) else [ctx]
+    if not ctxs:
+        return None
+    total = ctxs[0].calibrate()
+    for c in ctxs[1:]:
+        total = total + c.calibrate()
+    return total
+
+
+def _build_calibrated_losses(
```
HAOCHENYE: I found a bug here. Both the `balancing loss context` and `zloss context` should have a `cat` method implemented, similar to the `LMHeadLossContext`. Otherwise, under `micro_batch` conditions, it's impossible to calculate the aux loss correctly based on the whole.

HAOCHENYE: Additionally, I would like `AuxLoss` to implement a `calibrate` method, so that manual calls here are no longer necessary.

PR #1969, opened 2026-07-20
[Refactor] Drop non-global aux-loss averaging mode

HAOCHENYE:
Stacked on #1968.

Removes the `balancing_loss_global_average` / `z_loss_global_average` config switches and their non-global-average branch. That mode computes a full per-rank loss with **local** denominators, which is incompatible with reduce-sum — its replicated router/gate gradient would be inflated by `world_size` under SUM reduction. The `dist.is_initialized()`-False single-process path is kept as the world-size-1 case (numerically identical to the global branch at W=1).

**Breaking config change**: removes two config fields (`balancing_loss_global_average`, `z_loss_global_average`).

### DIFF
--- a/tests/model/test_reduce_sum_grad.py
+++ b/tests/model/test_reduce_sum_grad.py
@@ -1,0 +1,766 @@
+from contextlib import contextmanager
+
+import torch
+import torch.distributed as dist
+import torch.nn as nn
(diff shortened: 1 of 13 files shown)


HAOCHENYE:
Folded into #1968 (single squashed commit). The non-global aux-loss mode removal is now part of that PR. Closing.

PR #1972, opened 2026-07-21
[Test] Add backward coverage for DSA sparse_attn kernel and assembly

HAOCHENYE:
## Motivation

Review of the dsv4 branch's tests against the `submodule_tests` conventions surfaced two backward-coverage gaps in the DSA suite:

- **`sparse_attn` is a differentiable production kernel** (pure-PyTorch gather + softmax with autograd support), but its unit test only covered forward (`test_full_topk_matches_dense`, sink, masked-`-1`, shape). The convention mandates forward **and** backward vs a reference for an op/kernel. Whole-model parity does not close this: parity runs the eager `_sparse_attn_hf_parity` path, not the production gather kernel, so the production forward/backward is otherwise untested.
- **The DSA assembly smoke was forward-only.** The convention's smoke contract is "forward **and** backward run across a few shapes without error"; no `.backward()` was exercised on the DSA block anywhere.

## Changes

- `test_sparse_attn.py`: add `test_backward_matches_dense` — compares grads on `q` / `kv` / `attn_sink` against the existing dense reference at full top-k (relative tolerance for gather-vs-dense fp32 reduction reordering).
- `test_dsa.py`: add `test_backward_smoke` parametrized over `compress_ratio` ∈ {0, 4} (sliding + compressed-sparse paths) — runs the full hc_pre → projections → sparse_attn → grouped O-LoRA backward and asserts finite, non-zero grads on the always-in-path `wo_b` and on the hidden input.

## Test Plan

```
pytest tests/module/dsa/test_sparse_attn.py tests/module/dsa/test_dsa.py::TestDeepSeekSparseAttention
# 15 passed
```

CPU-only, native backend; no GPU required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/module/dsa/test_dsa.py
+++ b/tests/module/dsa/test_dsa.py
@@ -242,6 +242,34 @@ class TestDeepSeekSparseAttention:
             atol=0.0,
         )
 
+    @pytest.mark.parametrize("compress_ratio", [0, 4])
+    def test_backward_smoke(self, compress_ratio: int) -> None:
+        # Assembly-level smoke: the DSA block wires hc_pre → q/kv projections →
+        # (compressor/indexer) → sparse_attn → grouped O-LoRA, and the whole
+        # chain must run backward without error, not just forward. Correctness
+        # of each piece is owned by the op/module tests; this guards the wiring
+        # so a shape or dtype regression in the backward graph fails here rather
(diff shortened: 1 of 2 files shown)


HAOCHENYE:
Closing: applying these test additions directly to the dsv4 working tree for local review instead of a stacked PR.

PR #1973, opened 2026-07-21
[Test] Add intent base classes and register pytest markers

HAOCHENYE:
## What

Establishes the shared **test architecture baseline** for the dsv4 suite that the per-file test agents rebase onto.

### Intent base classes (`xtuner/_testing/intent.py`)

A submodule test's guarantee is now part of its **type**, visible at review before reading the body:

| Intent | Base class | Standard |
|---|---|---|
| parity — reproduce HF | `ParityTest` | strict `torch.equal`, no tolerance |
| correctness — match an oracle | `CorrectnessTest` | class-level `atol`/`rtol` (fp32 default `1e-3`, widened per bf16 subclass) |
| smoke — runs & stays finite | `SmokeTest` | outputs finite; no oracle |

Pure mixins, so a distributed test composes intent with the process base: `class TestFooCorrectness(CorrectnessTest, DeterministicDDPTestCase)`. Check standards are kept in separate methods on purpose — the class a test extends *is* the review signal, so they are deliberately not folded into one parametrised helper.

**"parity" means bitwise-against-HF only.** Backend agreement (Triton vs native), padded-vs-native, and distributed equivalence differ in the last bits → they are `CorrectnessTest`, not parity. Existing `…Parity` classes doing a tolerance compare between two backends should be reclassified to `…Correctness`/`…Equivalence`.

### Marker registration (`pyproject.toml`)

Adds `[tool.pytest.ini_options]` registering `gpu` / `npu` / `smoke` with `--strict-markers`. Removes the UnknownMark warning on `@pytest.mark.gpu` (the only custom marker used repo-wide) and turns a future typo'd marker into an error. Device markers gate where a test runs (`-m "not gpu"`); `smoke` doubles as a fast-subset selector.

### Skill (`submodule_tests`)

Defines the three intents + base classes, the directory/file/class/test naming hierarchy (subsystem / submodule / intent / contract), the class-docstring convention (under test / oracle / why enough), and the marker guidance.

## Test Plan

- `ruff check` / `ruff format --check` clean on new files.
- Base classes exercised: parity rejects a 1e-6 diff, correctness honors class tolerance + per-subclass override, smoke catches NaN.
- `pytest --markers` lists the three; `-m "not gpu"` deselects gpu-marked tests with no UnknownMark; `-m smoke` selects only the smoke class.

## Notes for rebasing agents

- Base branch is `dsv4`; this commit is meant to sit under the per-file test changes.
- When adopting: pick one intent base class per test class, rename `…Parity` backend/pad/distributed classes to correctness, and move `@pytest.mark.smoke` onto smoke classes.

### DIFF
--- a/.claude/skills/submodule_tests/SKILL.md
+++ b/.claude/skills/submodule_tests/SKILL.md
@@ -1,0 +1,282 @@
+---
+name: submodule_tests
+description: >
+  Decide what tests a single XTuner submodule needs before it is trusted in a
+  model. Use this when adding or reviewing a new operator, module, attention
+  helper, weight-mapping, or multi-backend kernel under `xtuner/v1` — especially
+  while porting a model (see `add_hf_model`), where each new piece must be
+  validated in isolation before whole-model parity. Maps each kind of submodule
+  to the specific tests it owns (forward/backward vs a readable reference, packed
+  correctness, pad safety, backend agreement, distributed equivalence, weight
+  loading), points at the canonical benchmark test for each, and says when a
+  piece should be left to parity instead of unit-tested. Defines the three test
+  intents and their base classes (`ParityTest` / `CorrectnessTest` / `SmokeTest`
+  in `xtuner._testing`), the directory/file/class/test naming hierarchy, and the
+  device/smoke pytest markers.
+---
+
+# What tests does this submodule need?
+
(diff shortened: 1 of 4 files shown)


HAOCHENYE:
Closing per request; not opening PRs unprompted.

PR #1994, opened 2026-07-31
docs: fix example config paths in English installation guide

latent-9:
The English installation guide references example configs that **404**:

- `examples/v1/sft_qwen3_tiny.py` (lines 71, 94)
- `examples/v1/sft_intern_s1_tiny_config.py` (lines 106, 270)

Both configs actually live under `examples/v1/config/` — `examples/v1/` contains only `config/` and `scripts/`. Running the documented command as-is fails at `Config.fromfile(...)` with `FileNotFoundError`.

The **Chinese** installation guide (`docs/zh_cn/get_started/installation.md`) already uses the correct `examples/v1/config/...` paths, so this simply brings the English doc back in line.

Docs-only, no functional change.

### DIFF
--- a/docs/en/get_started/installation.md
+++ b/docs/en/get_started/installation.md
@@ -68,7 +68,7 @@ Having problems running? Check out the [FAQ](faq)
 :caption: dense model fine-tuning example
 :linenos:
 
-torchrun  xtuner/v1/train/cli/sft.py --model-cfg examples/v1/sft_qwen3_tiny.py --chat_template qwen3 --dataset tests/resource/openai_sft.jsonl
+torchrun  xtuner/v1/train/cli/sft.py --model-cfg examples/v1/config/sft_qwen3_tiny.py --chat_template qwen3 --dataset tests/resource/openai_sft.jsonl
 ```
 
 After successful execution, the log is as follows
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
Verified against the tree rather than eyeballing the paths, and the change is right.

`examples/v1/` on `main` contains only two entries, both directories:

```
dir  config
dir  scripts
```

so nothing at `examples/v1/*.py` resolves, and the commands in the English installation guide fail for anyone copying them. The corrected targets are both present:

```
examples/v1/config/  ->  19 files, including
                         sft_qwen3_tiny.py
                         sft_intern_s1_tiny_config.py
```

The detail that makes this straightforward to accept is that `docs/zh_cn/get_started/installation.md` already carries the corrected paths:

```
torchrun  xtuner/v1/train/cli/sft.py --model-cfg examples/v1/config/sft_qwen3_tiny.py ...
torchrun xtuner/v1/train/cli/sft.py --config examples/v1/config/sft_intern_s1_tiny_config.py
```

So the two translations had drifted and this brings the English one back in line rather than proposing a new convention. Worth mentioning in the description, since "which one is correct" is the first question a reviewer has on a path fix, and the Chinese doc answers it.

Both the command lines and the surrounding prose references are updated in the same diff, which is the part that usually gets half-done. Someone reading "modify the `num_hidden_layers` parameter in ..." needs the same corrected path as the command above it.

Looks complete for this file.

latent-9:
Closing this typo-only change to keep the tree clean.

PR #2005, opened 2026-08-10
[Fix] Fix Ray generate concurrency group metadata

YanhuiDua:
## Summary
- Place `@ray.method` **closest to** the original `generate` method in both `RolloutController` and `RolloutWorker`.
- Keep `@trace_rollout_endpoint` as the outer wrapper.

## Problem
`trace_rollout_endpoint` wraps the decorated method with `functools.wraps`, which sets `wrapper.__wrapped__` to the original function.

With the previous decorator order:

```python
@ray.method(concurrency_group="generate")
@trace_rollout_endpoint(...)
async def generate(...):
    ...
```

@ray.method attached __ray_concurrency_group__ to the outer trace wrapper.When Ray builds actor method metadata, it calls inspect.unwrap(method) before reading __ray_concurrency_group__. Ray therefore inspected the original function, where the concurrency group attribute was absent.

As a result, both RolloutController.generate and RolloutWorker.generate were missing from Ray's concurrency group method mapping and executed in the default concurrency group.

### DIFF
--- a/docs/sphinx_patch/mock_obj.py
+++ b/docs/sphinx_patch/mock_obj.py
@@ -106,9 +106,18 @@ _orig_mockmod_getattr = mock._MockModule.__getattr__
 _orig_mockobj_getattr = mock._MockObject.__getattr__
 
 
+def _mock_ray_method(*args: Any, **kwargs: Any) -> Any:
+    """Keep ``ray.method`` as a pass-through decorator during autodoc imports."""
+    if len(args) == 1 and callable(args[0]) and not kwargs:
+        return args[0]
+    return lambda func: func
+
+
 def _patched_mockmod_getattr(self, name: str) -> Any:
     if name == "__version__":
         return _MOCK_VERSION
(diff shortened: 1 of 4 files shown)


YanhuiDua:
@claude review

docs/sphinx_patch/mock_obj.py
```diff
@@ -106,9 +106,18 @@ def _make_subclass(name: str, module: str, superclass: Any = _MockObject,
 _orig_mockobj_getattr = mock._MockObject.__getattr__
 
 
+def _mock_ray_method(*args: Any, **kwargs: Any) -> Any:
```
YanhuiDua, line 109: 调整装饰器顺序后，Sphinx mock 的 ray.method 会先将异步函数替换为 mock 对象并触发 trace 校验失败，因此将 docs 中的 ray.method mock 改为透传装饰器以恢复构建且不影响运行时行为。

jayhenry (approved):

PR #2016, opened 2026-08-12
[CI] Split Claude review analysis and publication

jayhenry:
## Root Cause

The review action used one 45-turn session for both investigation and publication. On a large PR, Claude started four background agents and was still collecting evidence when the hard limit terminated the session, so it posted neither inline comments nor a top-level summary. A static prompt could describe the budget but could not inject a live turn countdown.

The failed run also recorded 13 permission denials. Nine came from the background agents; the remaining four repeatedly tried to save the full PR patch through temporary-file redirection. In addition, `allowed_non_write_users: "*"` implicitly enabled the action subprocess-isolation mode even though this job already relies on runner isolation.

## Fix

- Split review into a 35-turn analysis phase and a 10-turn publication phase that resumes the same Claude session.
- Recover the session ID from the execution log when the expected analysis `max_turns` result prevents the action from exposing it directly.
- Remove `Agent` from the visible tool set and explicitly disallow it in both phases.
- Fetch the PR ref before invoking Claude and direct analysis to scoped `git diff` / `git show` calls without temporary files, redirection, loops, or compound commands.
- Opt out of the action-added subprocess isolation and retain the existing Bash allowlist and explicit deny rules.
- Restrict the publication phase to GitHub comment operations and publish the top-level summary before inline comments.

## Test Plan

- Parsed `.github/workflows/claude-general.yml` with PyYAML.
- Verified the two turn budgets, resume wiring, tool visibility, and `Agent` denial with configuration assertions.
- Ran actionlint 1.7.12. It reports only the same two pre-existing `pull_request_review_comment` activity-type warnings present on `upstream/main`; this change adds no new warnings.
- Ran `git diff --check`.

### DIFF
--- a/.github/scripts/prepare-claude-review-inputs.sh
+++ b/.github/scripts/prepare-claude-review-inputs.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ "$#" -ne 2 ] || [[ ! "$1" =~ ^[1-9][0-9]*$ ]]; then
+  echo "Usage: $0 <pr-number> <output-directory>" >&2
+  exit 2
+fi
+
+: "${GH_TOKEN:?GH_TOKEN must be set}"
+: "${GITHUB_REPOSITORY:?GITHUB_REPOSITORY must be set}"
+
+pr_number=$1
+output_dir=$2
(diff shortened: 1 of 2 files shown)


jayhenry:
@claude approve if ok

YanhuiDua (approved):

PR #2030, opened 2026-08-17
feat(muon): split fused QKV projections

RangiLyu:
## Motivation

Muon should orthogonalize the logical projections inside a fused QKV weight independently.

Treating a fused QKV weight as a single matrix couples unrelated Q, K, and V updates during Newton-Schulz orthogonalization and applies learning-rate scaling based on the combined matrix shape.

This PR uses the `get_muon_split_sizes()` interface introduced in #2001 to expose the logical row blocks of fused QKV parameters.

## Changes

- Add Muon split metadata for GatedDeltaNet `in_proj_qkv.weight`:
  - Q: `key_dim`
  - K: `key_dim`
  - V: `value_dim`
- Add Muon split metadata for Qwen3-VL Vision `qkv.weight`:
  - Q: `dim`
  - K: `dim`
  - V: `dim`
- Keep the forward computation and parameter layout unchanged.

The Muon optimizer configuration discovers these mappings through the module-level `get_muon_split_sizes()` interface and processes each logical projection independently.

### DIFF
--- a/xtuner/v1/model/compose/qwen3_vl/modeling_vision.py
+++ b/xtuner/v1/model/compose/qwen3_vl/modeling_vision.py
@@ -1,5 +1,5 @@
 import os
-from typing import List, Callable
+from typing import List, Callable, cast
 from xtuner.v1.ops.act_fn import get_act_fn
 import torch.nn as nn
 import torch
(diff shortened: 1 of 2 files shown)


CyCle1024:
@RangiLyu Is it necessary to use per head Muon? This PR just splits Q K V into 3 matrix orthogonalization.

RangiLyu:
> @RangiLyu Is it necessary to use per head Muon? This PR just splits Q K V into 3 matrix orthogonalization.

Per-head Muon split was introduced as a training trick in GLM-5. However, this PR targets Qwen3.5 and GDN, where the main issue is different: projections that are logically independent are fused into a single QKV weight and therefore orthogonalized together by Muon. The purpose of this change is to restore the original projection boundaries.

As I know, Qwen3.5 and GDN-related works did not suggest that Muon should be applied per head. Moreover, recent work https://arxiv.org/html/2605.08933v1 shows that finer-grained splitting is not always better, and the optimal orthogonalization granularity may depend on the training stage and gradient geometry.

So I would prefer to keep this PR focused on fixing the fused-projection issue, rather than introducing per-head Muon as a default behavior without architecture-specific evidence.

CyCle1024 (approved):

PR #2040, opened 2026-08-24
[Feature] Instrument sandbox agents and SessionServer tracing

matrix72c:
## Summary

Add OpenTelemetry instrumentation to the rollout, sandbox, and SessionServer paths so a request trace remains connected across Ray actors, sandbox stages, and the proxy lifecycle.

## Motivation

The existing rollout and SessionServer paths already expose the live boundaries needed for tracing, but their OTel context was not consistently propagated across Ray calls or represented in sandbox and proxy spans. This made cross-process request traces incomplete even though the underlying training behavior was correct.

## Implementation

- Propagate the active OTel context through Ray `generate_group` calls for a single `RolloutState` or a concrete `list`/`tuple` group, while restoring internal carrier and call-chain fields after the call.
- Bridge the existing sandbox stage timer to live OTel spans with stable names and a bounded attribute schema. Legacy JSONL enter/exit records and timing semantics remain unchanged.
- Trace the SessionServer proxy lifecycle as `prepare_request`, `backend_roundtrip`, and `record_response` under one request span. Incoming aiohttp headers are used as the parent carrier, and the backend carrier is injected inside the roundtrip span.

## Compatibility

- No training behavior or request/response payload format changes.
- No prompt, response body, or raw header attributes are exported.
- No duplicate duration/ok telemetry is added; OTel owns duration and logical failures use `error=true`.
- This PR instruments existing live call sites only. It does not add synthetic trace reconstruction, trajectory polling, timeline storage, or a flush API.

## Tests

- Focused rollout propagation, nested cleanup, sandbox bridge, SessionServer, and mixed-case W3C carrier tests.
- Existing `SingleTurnAgentLoop` and producer compatibility tests: 29 passed.
- `ruff check` on all changed Python files.
- `compileall` on changed modules and tests.
- `git diff --check` and final diff review against `upstream/main`.

### DIFF
--- a/tests/rl/test_trace.py
+++ b/tests/rl/test_trace.py
@@ -3,12 +3,12 @@ import json
 import os
 import subprocess
 import sys
+import tempfile
 import types
 import unittest
+import unittest.mock
 from contextlib import contextmanager
 from pathlib import Path
-from tempfile import TemporaryDirectory
-from unittest import mock
 
 
 def _run_trace_utils(repo_root: Path, command: str) -> dict:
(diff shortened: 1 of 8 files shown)


YanhuiDua:
@matrix72c XTuner中trace功能设计的目标为不保留打点代码在框架中，由xtuner-trace SKILL辅助打点，否则对于框架侵入式修改过大，建议这个PR修改在你的开发分支中维护

PR #2041, opened 2026-08-24
[Feature] Support external OTLP collector for trace runtime

matrix72c:
## Summary

Allow XTuner processes to export spans to an externally managed, network-reachable OTLP collector while preserving the existing driver-local collector and viewer defaults.

## Motivation

The default trace path introduced in #1946 uses a collector on the driver's loopback address. In a multi-pod deployment, a remote Ray actor resolves `127.0.0.1` to its own pod, so it cannot reach the driver's collector and its spans are missing from the JSONL and Jaeger traces.

This PR allows the producers to use a routable collector Service instead:

```text
driver / Ray actors in multiple pods
        | OTLP/gRPC
        v
external collector Service
        |-- file exporter --> shared traces.jsonl --> XTuner viewer
        `-- exporter ------> Jaeger or another backend
```

## Configuration

```python
TraceConfig(
    enabled=True,
    external_otlp_endpoint="http://otel-collector.namespace.svc:4317",
    external_trace_jsonl_path="/shared/traces/traces.jsonl",
    xtuner_viewer_enabled=True,
)
```

- `external_otlp_endpoint` is propagated to Ray child processes so every process exports to the same reachable collector.
- The external collector is deployed, configured, started, and stopped outside XTuner.
- `external_trace_jsonl_path` only tells the driver-side viewer where to read the collector-owned JSONL file; it does not configure or start the collector and is not propagated to Ray child processes.
- When the viewer is enabled, the path must refer to the same underlying file written by the external collector and must be visible from the driver, for example through a shared volume.
- The viewer filters a shared JSONL file by `service.name` and XTuner `run.id`, so the file may contain spans from multiple runs.
- If the external collector only exports to Jaeger or another backend, leave `xtuner_viewer_enabled=False`; no JSONL path is required.
- Viewer JSONL loading is deferred until the first request so the external collector may create the file after the viewer process starts.

## Compatibility

When `external_otlp_endpoint` is unset, local collector startup, per-run `traces.jsonl` output, viewer startup, and shutdown behavior remain unchanged. External collectors are never started or stopped by XTuner.

## Tests

- 8 focused tests covering local defaults, external collector lifecycle, endpoint propagation to Ray children, collector ownership of JSONL, deferred viewer loading, shared JSONL viewer configuration, and invalid configuration combinations.
- `ruff check`.
- `compileall`.
- `git diff --check`.

### DIFF
--- a/recipe/trace/viewer/server.py
+++ b/recipe/trace/viewer/server.py
@@ -181,7 +181,6 @@ def start_rollout_trace_viewer(
         load_base_payload,
         source_signature=current_source_signature,
     )
-    payload_cache.get(train_step)
 
     class Handler(http.server.BaseHTTPRequestHandler):
         def do_GET(self) -> None:
--- a/tests/rl/test_trace.py
+++ b/tests/rl/test_trace.py
@@ -4,6 +4,8 @@ import subprocess
 import sys
 import unittest
 from pathlib import Path
+from tempfile import TemporaryDirectory
+from unittest import mock
 
 
 def _run_trace_utils(repo_root: Path, command: str) -> dict:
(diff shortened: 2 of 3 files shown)


YanhuiDua:
最简单的修复是否是 driver 提供的endpoint为其真实IP，而非 127.0.0.1，这个问题能否一起帮忙修复下？支持external collector 本身没问题。

YanhuiDua:
@claude review

PR #2044, opened 2026-08-25
[Fix] Align FP8 RL train and rollout weight quantization

jayhenry:
## Summary

- Add the default-off `XTUNER_RL_FP8_QUANTIZE_IN_BF16` switch. When enabled, tilewise FSDP scale precomputation and FP8 casting quantize the same BF16-rounded weight values exported to rollout, while keeping the FP32 optimizer master unchanged.
- Pass runtime FP8 metadata to direct rollout weight conversion and cover BF16 HF/direct-FP8 export behavior in the existing WeightIterator tests.
- Include the switch in the standard trainer environment report and print all `XTUNER*` variables in the RL trainer runtime environment report.

The legacy FP32-master quantization path remains the default, and the existing FP8 engine loss references are unchanged.

## Validation

- Pre-commit hooks: ruff, ruff-format, mypy, codespell, docformatter, and pydantic checks passed.
- `tests/engine/test_moe_train_engine_float8.py`: 7 passed with the original loss references.
- CUDA float8/WeightIterator/HF-save suite: 22 passed.

### Train/rollout mismatch before and after the fix

Both columns use the two-step Qwen3.5-35B-A3B GSM8K FP8 setup with real optimizer updates.

| Metric | Before fix: FP32 train quantization / BF16 rollout export | With fix: BF16-aligned quantization |
|---|---:|---:|
| Step 1 `k3_kl` | `3.414027e-4` | `3.324241e-4` |
| Step 2 `k3_kl` | `8.691058e-3` | `3.729346e-4` |
| Step 2 / Step 1 | `25.457x` | `1.121864x` |
| Absolute growth after the first update | `8.349655e-3` | `4.051050e-5` |
| Step 2 mismatch reduction | baseline | `95.71%` |

The fix keeps the mismatch at the initial `1e-4` scale instead of jumping to `1e-2`; both fixed steps had non-zero gradient norms.

### DIFF
--- a/xtuner/v1/train/rl_trainer.py
+++ b/xtuner/v1/train/rl_trainer.py
@@ -655,6 +655,15 @@ class BaseRLTrainer:
 
         env_path = log_dir / "env.json"
         environment_variables = dict(os.environ)
+        xtuner_environment = {
+            key: value for key, value in sorted(environment_variables.items()) if key.startswith("XTUNER")
+        }
+        log_str = "\n============XTuner RL Training Environment============\n"
+        for key, value in xtuner_environment.items():
+            log_str += f"{key}: {value}\n"
+        log_str += "======================================================"
+        self.logger.info(log_str)
+
         infer_engine_version = get_rollout_engine_version()
         environment_variables.update(infer_engine_version)
         with env_path.open("w") as f:
--- a/tests/rl/test_weight_iterator.py
+++ b/tests/rl/test_weight_iterator.py
@@ -6,6 +6,9 @@ import torch
 import torch.distributed as dist
 from torch import nn
 
+from xtuner.v1.float8 import Float8Config, ScalingGranularity
+from xtuner.v1.float8.fsdp_utils import WeightWithDynamicTilewiseFloat8CastTensor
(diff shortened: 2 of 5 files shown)


xtuner/v1/train/rl_trainer.py
```diff
@@ -655,6 +655,15 @@ def _save_runtime_environment(self, log_dir: Path) -> None:
...
         env_path = log_dir / "env.json"
         environment_variables = dict(os.environ)
+        xtuner_environment = {
+            key: value for key, value in sorted(environment_variables.items()) if key.startswith("XTUNER")
+        }
+        log_str = "\n============XTuner RL Training Environment============\n"
+        for key, value in xtuner_environment.items():
+            log_str += f"{key}: {value}\n"
+        log_str += "======================================================"
+        self.logger.info(log_str)
```
YanhuiDua, line 665: 现在会保存env.json记录所有环境变量，这个还需要吗？

YanhuiDua (approved):

xtuner/v1/train/rl_trainer.py
jayhenry: 这里类似 sft 只打印了 XTUNER_xxx 变量

PR #1934, opened 2026-06-24
[Feature] Support multi-segment trainable rollout in sandbox agent loop

Harold-lkk:

PR #2045, opened 2026-08-25
[Fix] Deduplicate agent artifacts across rollout segments

matrix72c:
## Summary

This PR fixes the repeated ownership and serialization of session-level agent artifacts after a sandbox rollout is split into multiple trainable segments.

- Keep the complete artifacts on segment 0 instead of deep-copying them into every segment.
- Preserve artifacts on the single state returned by failed train rollouts and eval rollouts.
- Keep per-segment messages and tools, and expose the session ID, segment index, and segment count in trajectory rows.
- Stream rows into the existing trajectory JSONL instead of staging all serialized rows in memory.

## Root Cause

#1926 added complete agent artifacts to RolloutState for rollout observability. #1934 later added multi-segment training by deep-copying the base RolloutState once per trainable segment.

The artifacts were attached before that deep copy. A session with N trainable segments therefore retained N independent copies of the same session-level payload. The trajectory writer then materialized every row in memory and serialized the repeated artifacts into every segment row, amplifying both CPU memory and filesystem I/O by the segment count.

The model training path does not consume agent artifacts. By the time a segment reaches training, the required input IDs, labels, log probabilities, routed experts, and reward have already been materialized.

## Fix

For successful training rollouts, build every segment from a base state that does not contain agent artifacts, then attach one JSON-safe artifacts payload to segment 0 after segment construction. Failed train rollouts and eval rollouts still return one state and retain their complete artifacts.

The trajectory writer continues to produce one train_rollout or eval_rollout JSONL file. The summary remains the first line, followed by segment rows written one at a time. Each row includes session_id, segment_index, and segment_count. The artifacts field remains in its existing location: it contains the complete payload on segment 0 and is null on later segments.

This keeps ownership explicit at the rollout source. The writer does not maintain a seen-session cache, create a companion file, or silently repair duplicated input states.

## Impact

- Training tensors, rewards, advantages, messages, tools, and segment ordering are unchanged.
- Complete artifacts remain available once per session for debugging and observability.
- Existing non-agent rollouts keep null agent metadata.
- No configuration option, new session abstraction, ObjectRef store, compression layer, or companion-file lifecycle is introduced.

Consumers that previously read agent.artifacts from every segment should read it from segment 0, identified by session_id and segment_index.

## Reproduction

The issue was observed in a train_rollout_1.jsonl file with 1,275 segment rows and a total size of 68.8 GB. In one sampled session:

| Metric | Value |
| --- | ---: |
| Trainable segments | 22 |
| One serialized artifacts payload | 3,358,105 bytes |
| Artifacts bytes written before this fix | 73,878,310 bytes |
| Artifacts bytes written after this fix | 3,358,105 bytes |

The sampled artifacts payload was identical across all 22 segment rows. Routed experts were not serialized in the JSONL and are unrelated to this issue.

## Test Plan

- Trajectory ownership and logging regression tests: 8 passed.
- Related colocated and disaggregated trainer lifecycle tests: 6 passed.
- Upstream GitHub Actions lint workflow: passed.
- Local codespell, docformatter, pyupgrade, Ruff lint, Ruff format, and git diff checks: passed.

The full repository unit-test matrix is left to upstream CI.

## Out of Scope

This PR does not change routed-expert storage, trace-store lifetime, replay-buffer grouping, JSON compression, or the contents of the artifacts payload itself.

### DIFF
--- a/tests/rl/test_trajectory_logging.py
+++ b/tests/rl/test_trajectory_logging.py
@@ -1,0 +1,322 @@
+import copy
+import json
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+from xtuner.v1.data_proto.rl_data import RolloutState, Status
+from xtuner.v1.train.rl_trainer import BaseRLTrainer
+
+
+with patch.dict(
+    sys.modules,
+    {
+        "lagent": MagicMock(),
+        "lagent.utils": MagicMock(),
+        "lagent.utils.rate_limiter": MagicMock(),
+    },
+):
+    from xtuner.v1.rl.agent_loop.sandbox_agent_loop.agent_in_sandbox_loop import AgentInSandboxLoop
+    from xtuner.v1.rl.agent_loop.sandbox_agent_loop.schemas import AgentRolloutItem, RolloutStatus
+
+
+class TestSandboxArtifactOwnership(unittest.IsolatedAsyncioTestCase):
+    @staticmethod
+    def _make_state() -> RolloutState:
+        return RolloutState(
+            group_id=3,
+            rollout_id=7,
+            session_id=12345678901234567890,
+            message=[{"role": "user", "content": "test"}],
+            num_tokens=1,
+            extra_fields={},
+        )
+
+    @staticmethod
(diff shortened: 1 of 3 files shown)


YanhuiDua:
@claude review

YanhuiDua (approved):
LGTM

PR #2047, opened 2026-08-26
fix: make Muon gradient clipping configurable

Denny991:
## Summary

This PR makes the Muon gradient clipping policy configurable.

`adamw_only` remains the default policy. The historical `qwen3-5-sft-ep2-mtp4-vl` E2E case explicitly uses `all` to align with its existing baseline.

## Problem

After `d1ecfdd`, the Qwen3.5 E2E case showed larger `grad_norm` fluctuations and persistent loss differences, while LR and token counts remained identical.

The existing baseline was generated with all-group clipping, but `d1ecfdd` changed the default to AdamW-only clipping.

Failing run: https://github.com/InternLM/xtuner/actions/runs/32438896110

## Changes

- Add `MuonConfig.clip_grad_mode`:
  - `adamw_only`: clip only AdamW parameter groups.
  - `all`: clip both Muon and AdamW parameter groups.
- Keep `adamw_only` as the default.
- Set `clip_grad_mode="all"` only for the historical Qwen3.5 E2E case.
- Add tests for both policies and optimizer state loading.

## Validation

- 8-GPU A/B confirmed that the clipping policy caused the loss and `grad_norm` differences.
- Explicit `all` reproduced the pre-change behavior within observed run-to-run noise.
- Lint passed.
- Unit test and 16-GPU E2E are pending.

Related change: https://github.com/InternLM/xtuner/pull/2001

### DIFF
--- a/xtuner/v1/config/optim.py
+++ b/xtuner/v1/config/optim.py
@@ -78,6 +78,10 @@ class MuonConfig(OptimConfig):
     enable_all2all: Annotated[
         bool, Parameter(help="Allow all-to-all comm strategy; set False on topologies where all-to-all is unreliable")
     ] = True
+    clip_grad_mode: Annotated[
+        Literal["all", "adamw_only"],
+        Parameter(help="Gradient clipping policy: clip all parameters or only the AdamW parameter groups"),
+    ] = "adamw_only"
 
     def build(self, model):
         trainable_parameters_names = model.trainable_parameters()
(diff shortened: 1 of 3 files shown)


xtuner/v1/config/optim.py
```diff
@@ -78,6 +78,11 @@ class MuonConfig(OptimConfig):
     enable_all2all: Annotated[
         bool, Parameter(help="Allow all-to-all comm strategy; set False on topologies where all-to-all is unreliable")
     ] = True
+    # Preserve the pre-selective-clipping behavior unless a model explicitly opts in.
+    clip_grad_mode: Annotated[
+        Literal["all", "adamw_only"],
+        Parameter(help="Gradient clipping policy: clip all parameters or only the AdamW parameter groups"),
+    ] = "all"
```
jayhenry: 默认改为更正确的 `adamw_only`， 在端到端测试中显式设置为 all 来对齐baseline

Denny991: > 默认情况下是更正确的`adamw_only`，在最终测试中显着方式设置为所有的横向基线

好，已经修改了

jayhenry (approved):

PR #2061, opened 2026-09-01
[Fix] Classify SessionServer endpoints and fail closed on trace errors

matrix72c:
## Summary

`SessionServer` is a catch-all HTTP proxy, but only two requests represent model generations: `POST /v1/messages` and `POST /v1/chat/completions`.

This change makes endpoint policy explicit before request/response processing and makes trace-enabled generations fail closed when the response cannot be recorded. Auxiliary API calls remain transparent, while the client-visible result stays consistent with the training trace stored by `SessionServer`.

The response-reliability changes previously proposed in #2059 are consolidated here so endpoint and trace policy are reviewed in one PR.

## Problem

The proxy cannot safely infer the endpoint from the JSON body. For example, `POST /v1/messages/count_tokens` contains `messages` but is not a model generation. Treating it as one can invoke `on_request`, inject generation-only fields, filter tools, or invoke `on_response` unexpectedly.

For trace-enabled generations, malformed JSON, non-object response bodies, incomplete SSE streams, downstream disconnects, or a failing `on_response` hook can leave the trace store without a complete assistant turn. Returning a normal success response in that situation allows the caller to observe a turn that cannot be used for training.

## Implementation

- Add a small private `_is_generation_endpoint(method, path)` predicate. It uses the normalized method and path, ignores query strings, and recognizes only `POST /v1/messages` and `POST /v1/chat/completions` as generation endpoints.
- Keep format detection independent from generation classification, so `/v1/messages/count_tokens`, `/v1/messages/batches`, and future `/v1/messages/...` paths still receive the correct Anthropic headers.
- Keep endpoint policy explicit and local to `_handle_request`; no `ContextVar`, wrapper handler, shared enum, or cross-project abstraction is introduced.
- For non-generation JSON object requests, remove only the top-level SessionServer-owned `session_id`. Non-object bodies and objects without `session_id` remain byte-for-byte unchanged whenever possible.
- Derive response tracing from both the endpoint policy and the upstream status code. Non-`2xx` responses never enter the cleaner or trace hooks, and explicit upstream error envelopes are passed through without being wrapped again.
- For trace-enabled streams, retain the original upstream chunks and delay `[DONE]` until parsing and `on_response` complete. If parsing or the response hook fails, return a native `500` error or SSE error event without `[DONE]`.
- Continue draining the upstream stream after a downstream connection reset so a complete trace can still be collected.
- Handle stream preparation and connection-reset behavior directly in the request handler instead of adding a one-use helper.

Successful generation behavior, the public `SessionServer` API, and the existing `/v1/responses` `501` rejection remain unchanged. No LMDeploy code or management endpoint implementation is changed, and `fix/chunk-loss-detached-head-memory` is untouched.

## Tests

Added focused `aiohttp` fake-upstream coverage for:

- generation and non-generation method/path classification;
- transparent `count_tokens` and auxiliary endpoint forwarding with no generation hooks;
- successful training responses and delayed stream completion;
- evaluation requests that skip `on_response`;
- upstream JSON/SSE errors and all non-`2xx` responses remaining unchanged;
- malformed, non-object, incomplete, and untraceable `2xx` responses failing closed;
- response-hook failures producing native errors without `[DONE]`; and
- downstream disconnects before headers and during streaming while the upstream is still drained and traced.

Validation performed:

```text
Focused endpoint/trace suite: 25 passed
ruff check: All checks passed
python -m py_compile: passed
git diff --check: passed
```

### DIFF
--- a/tests/rl/test_session_server_endpoints.py
+++ b/tests/rl/test_session_server_endpoints.py
@@ -1,6 +1,5 @@
 import json
 import unittest
-from copy import deepcopy
 from unittest.mock import AsyncMock
 
 from aiohttp import web
@@ -10,21 +9,21 @@ from xtuner.v1.rl.rollout.session_server import (
     FMT_ANTHROPIC,
     FMT_OPENAI,
     SessionServer,
-    _classify_generation_endpoint,
     _detect_format,
+    _is_generation_endpoint,
 )
 
 
 class TestEndpointClassification(unittest.TestCase):
     def test_only_generation_endpoints_are_classified(self):
         cases = [
-            ("POST", "v1/messages", "anthropic_messages"),
-            ("POST", "/proxy/v1/messages?beta=true", "anthropic_messages"),
-            ("POST", "/v1/chat/completions", "openai_chat_completions"),
+            ("POST", "v1/messages"),
+            ("POST", "/proxy/v1/messages?beta=true"),
+            ("POST", "/v1/chat/completions"),
         ]
-        for method, path, expected in cases:
+        for method, path in cases:
             with self.subTest(method=method, path=path):
-                self.assertEqual(_classify_generation_endpoint(method, path), expected)
+                self.assertTrue(_is_generation_endpoint(method, path))
 
     def test_non_generation_paths_and_wrong_methods_are_not_classified(self):
         cases = [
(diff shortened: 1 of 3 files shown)


YifanHe-ailab:
@claude review

YifanHe-ailab:
@claude review

YifanHe-ailab:
@claude review

YifanHe-ailab (approved):

Issue #2025, opened 2026-08-14
[Bug] Routed-expert traces cause host OOM with offload enabled and CUDA OOM before SP split when disabled


matrix72c:
## Summary

In a multi-node MoE RLOO training run, `offload_rollout_routed_experts` only changes where a large set of rollout routing traces exhausts memory:

- With `offload_rollout_routed_experts=True`, learner host memory grows until Ray's 95% node-memory threshold is reached.
- With `offload_rollout_routed_experts=False`, the first learner step fails with CUDA OOM while moving `rollout_routed_experts` to the GPU, before sequence-parallel sharding is applied.

The current behavior suggests that rollout routing traces are materialized for too much of the learner batch at once, and that the non-offload path transfers the unsharded trace to every SP rank.

## Environment

The relevant code path is still present on current upstream `main` at commit [`575d7e0`](https://github.com/InternLM/xtuner/commit/575d7e058040baa7f609b3d5d3f397653877bc25).

Observed training topology/configuration:

- 3 nodes, 8 H200 GPUs per node
- 16 training workers and 8 rollout workers
- Sequence parallel size: 8
- Train batch size: 32 rollout groups
- RLOO repeat K: 4
- Maximum packed sequence length: 262,144
- The first 32 rollout groups expanded to approximately 2,390 learner samples

The run used a nearby checkout with an unrelated rollout-session lifecycle patch. The routing-trace code discussed below matches upstream `main`.

### `offload_rollout_routed_experts=True`: host OOM

The rollout phase completes and the learner prepares approximately 2,390 training batches. Host memory then crosses Ray's 95% node-memory threshold. Eight training workers on the affected learner node each use approximately 100 GiB of memory. Ray kills workers and the training step fails before the first optimizer update.

Ray object-store usage is comparatively small, so the dominant pressure appears to be worker heap rather than the object store.

### `offload_rollout_routed_experts=False`: CUDA OOM

The same workload reaches the first learner step, then fails in this path:

```text
TrainingWorker.fit()
  -> seq_ctx = data["seq_ctx"].to(DEVICE)
  -> SequenceContext.to()
  -> self.rollout_routed_experts.to(device)
  -> CUDA out of memory
```

At failure, the H200 had less than 200 MiB free and PyTorch attempted to allocate another approximately 640 MiB. Reserved-but-unallocated memory was negligible, so this looked like capacity exhaustion rather than fragmentation.

## Suspected cause

[...]

PR #2064, opened 2026-09-03
[Fix] Transfer routed experts release ownership

matrix72c:
## Motivation

When `enable_return_routed_experts=True`, an RL rollout can attach Ray `ObjectRef`s to a `RolloutState`. The same reference may either be created by the rollout worker or be borrowed from a `TraceStore` session. Before this change, `RolloutState` did not record which component owned the reference, while several cleanup helpers unconditionally freed references. This made retryable stale samples unsafe and could invalidate references that were still held by the `TraceStore`.

## Failure chain before this PR

The problematic path is easiest to see for a retryable stale sample:

1. A rollout group becomes stale in `ReplayBuffer` and is selected for retry.
2. The old lifecycle calls `reset_rollout_response()` to keep the prompt and clear the generated response.
3. `reset_rollout_response()` also calls `free_object_refs()` for every routed-experts `ObjectRef` it sees. It cannot distinguish a locally created rollout reference from a borrowed `TraceStore` reference.
4. For a `TraceStore` sample, the `Trie` still contains the same reference because the session has not been released yet. The reset therefore frees a shared object too early.
5. The retry keeps the existing trace/session lifecycle, and a later export or training read tries to fetch the reference from the `Trie`. The reference is already invalid, which can surface as an object-fetch failure or as routed-experts/sequence-length validation errors.

There were two related versions of the same ownership bug:

- Partial-rollout post-processing awaited its input routed-experts reference and then freed it unconditionally. A borrowed reference could therefore be destroyed while the `TraceStore` still owned it.
- Generic discard recursively freed references without an explicit release decision. Cleanup correctness depended on the caller knowing whether a session had already been released. In addition, the training-batch cleanup ran only after a successful prepare/fit sequence, so an exception could leave a `TraceStore` session alive.

Finally, replacing a value in the `TraceStore` trie did not release routed-experts references from the overwritten value, which could leak Ray objects during reroll/overwrite workloads.

## What changed

- Add `RolloutState.routed_experts_owner` with the values `"rollout"` and `"trace_store"`.
- Make `reset_rollout_response()` a pure state reset: it clears response fields and routed-experts fields, but never calls `ray.free`.
- Add `release_owned_routed_experts()` for explicit caller-owned release, and make `discard_rollout_state(..., release_refs=...)` opt in to releasing resources.
- Mark routed-experts references at every direct rollout producer (`RolloutWorker`, vLLM rollout parsing, and the VERL tool loop) and at every `TraceStore` export boundary.
- On retryable stale samples, call `release_owned_routed_experts()` unconditionally: `"rollout"` refs are freed, `"trace_store"` refs are only detached (they stay valid until their session is released), and untagged refs (restored from legacy checkpoints via `ray.put`) are also freed so they cannot leak.
- Make partial-rollout input release explicit. `RolloutWorker.generate()` opts in for direct rollout inputs; the handler default is non-releasing for safe reuse by other callers.
- Keep `TraceStore` session release as the final release point for its references, detach released routes before generic discard, and ensure the trainer performs session release in a `finally` block.
- On trie overwrite, park the replaced value instead of freeing its refs immediately: sibling `RolloutState`s may still borrow those refs, so the overwrite path cannot safely decide reachability. Parked refs are freed (deduplicated against the live tree) by the same session release that frees the trie.

## Ownership after this change

```text
Direct rollout:
  producer creates ref -> owner = "rollout"
  caller explicitly releases ref -> reset/discard clears state

TraceStore export:
  Trie/session owns ref -> RolloutState borrows it (owner = "trace_store")
  retryable reset only detaches it
  session release -> Trie frees the ref
```

The LMDeploy generation protocol is unchanged. Ownership is established at XTuner's producer/export boundaries, and old checkpoints remain loadable because the new field defaults to `None`.

## Impact

This prevents premature freeing of shared `TraceStore` references, makes direct-rollout cleanup explicit, and closes exception/overwrite cleanup gaps without introducing a global reference registry or changing stale reroll/session semantics.

## Tests

The following targeted checks pass locally:

- `ruff check` for all changed source and test files
- `ruff format --check` for all changed source and test files
- `python -m py_compile` for all changed source and test files
- RL state/rollout/producer/trajectory tests: **95 passed** in total
- Staleness ownership tests: **10 passed**
- Retryable replay ownership tests: **3 passed** (release / trace-store detach / untagged release)
- Rollout logic tests: **52 passed**
- Producer tests: **25 passed**
- Trajectory logging tests: **8 passed**
- TraceStore unit tests: **8 passed** (Ray local mode), including trie-overwrite non-freeing and session-release collection of overwritten refs

The full replay-buffer suite was also inspected; an existing async save/resume test does not complete in this shared test environment, so it was not used as a passing signal.

## Related issue

Related to [#2025](https://github.com/InternLM/xtuner/issues/2025). This PR fixes routed-experts `ObjectRef` lifetime leaks and premature frees on the XTuner RL path. It does not, by itself, bound learner-side materialization or change sequence-parallel transfer order; those memory-footprint issues remain separate follow-up work. The LMDeploy endpoint and generation protocol are unchanged.


YifanHe-ailab:
@claude review

matrix72c:
Addressed the review feedback in 4f3da9d5:

- **Trie overwrite premature free (W1) + hot-path cost (W2)**: accepted. `Trie.insert` no longer frees overwritten refs immediately — the trie cannot see refs still borrowed by sibling `RolloutState`s. Overwritten values are parked and freed by the session release (`Trie.release` stays the single release point), deduplicated against the live tree via a shared `seen` set. This also removes the full-trie scan + `model_dump` deep copies from the `on_response` hot path; the overwrite path is now O(1).
- **Owner-guard duplicated at call site (W4)**: accepted. `_apply_staleness_lifecycle` now calls `release_owned_routed_experts(item)` unconditionally; owner semantics live in one place (`rl_data.py`).
- **Untagged owner leak (W5)**: accepted with a safer rule. Only `"trace_store"` is protected now; refs with no owner tag (legacy-checkpoint restores re-`ray.put` by `_restore_nested_objectrefs`, no other borrower) are released instead of leaking. We did not adopt "treat unknown as rollout" blindly — borrowing must be explicitly declared.
- **Tests mocked the helper under test (W3)**: accepted. The replay-buffer cases now patch the outer boundary `ray_utils.free_object_refs` and assert real release/detach behavior; added an owner=None case. TraceStore cases assert overwrite does not free borrowed refs (`ray.get` still succeeds) and that session release collects parked refs with cross-tree dedup.

xtuner/v1/rl/rollout/worker.py
```diff
@@ -1031,6 +1032,7 @@ async def generate(self, rollout_state: RolloutState) -> RolloutState:
             if rollout_state.status == Status.FAILED:
                 error_msg = rollout_state.error_msg
                 status = rollout_state.status
+                release_owned_routed_experts(rollout_state)
```
YanhuiDua: 能否只修改reset_rollout_response 和 discard_rollout_state 内部的行为，这样就不用在每个调用点新增调用 release_owned_routed_experts 了

xtuner/v1/rl/rollout/utils.py
```diff
@@ -150,8 +149,13 @@ async def postprocess(
         status: Status,
         prompt_tokens: int,
         completion_tokens: int,
+        release_input_routed_experts: bool = False,
```
YanhuiDua: 为什么要增加这个参数，什么情况下 release_input_routed_experts = False

YanhuiDua:
长期设计看，ObjectRef 的生命周期管理仍然过于分散：ReplayBuffer、PartialRolloutHandler、RolloutWorker、TraceStore、TrainerWorker 和 RLTrainer 都需要理解引用所有权并参与清理，reset/discard 还需要通过 release_refs 等参数控制行为，后续很容易继续引入重复释放或泄漏问题。

**建议后续将 routed experts 抽象为独立的分布式 RoutedExpertsStore：**

- 每个节点部署 NodeStore，在本地保存实际 routed-experts shard 和对应 ObjectRef；
- 由统一的 Directory 维护 resource_id -> shard manifest 以及 owner/consumer lease；
- RolloutState、TraceStore 和 ReplayBuffer 只保存轻量 RoutedExpertsHandle(resource_id, shape, dtype, ...)，不直接保存和释放 ObjectRef；
- rollout 写入后创建 owner lease，训练消费前申请 consumer lease，session/group 完成后释放 owner lease；
- 当 owner 和 consumer lease 全部释放后，由 Directory 通知各 NodeStore 删除本地 shard；
- lease 应使用唯一 lease_id 集合维护，而不是裸整数引用计数，以保证 acquire/release 在 Ray task retry 下具备幂等性。

这样可以将所有物理释放操作收敛到 NodeStore，业务组件只管理 Handle 和 lease，并进一步删除 routed_experts_owner、release_owned_routed_experts()、discard_rollout_state(release_refs=...)、TraceStore 递归遍历 ObjectRef，以及 TrainerWorker 直接调用 free_object_refs() 等分散逻辑。

YanhuiDua:
@claude review

xtuner/v1/rl/rollout/worker.py
matrix72c: ok

xtuner/v1/rl/rollout/utils.py
matrix72c: handler 无法判断它是本进程产出的还是从 TraceStore 借来的，现在就一个调用方，也可以先简化掉。

YanhuiDua: 我建议这个简化掉，不要增加接口的复杂性，默认就是True的配置就好

xtuner/v1/rl/rollout/trace_store.py
```diff
@@ -173,6 +187,12 @@ def insert(self, key: str, value: Any) -> None:
                 node = node.children[key]
                 break
 
+        old_value = node.value
+        if old_value is not None and old_value is not value:
+            # A rerolled turn may overwrite an existing key.  The old value's
+            # refs may still be borrowed by RolloutStates, so park the value
+            # instead of freeing immediately; the session release frees it.
+            self._overwritten_values.append(old_value)
         node.value = value
```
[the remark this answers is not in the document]
matrix72c: Accepted in a4e94b16. The park strategy is gone: `Trie.insert` now simply overwrites `node.value = value` without freeing and without parking, and `_overwritten_values` is deleted.

Rationale, as suggested: dropping the trie's handle is enough — Ray's distributed reference counting reclaims the object once every holder (the trie, sibling `RolloutState` borrowers, in-flight `ray.get`) drops its own handle, so the object is neither leaked nor invalidated for borrowers. Parking the whole old `TokenizedSegment` instead pinned the routed-experts object plus token_ids/labels/logprobs in the store actor's memory and in the object store until session release, which grows without bound under reroll-heavy MoE sessions. The original motivation ("trie overwrite leaks refs") was wrong: an unreferenced-but-not-yet-counted-down object is not a leak.

`Trie.release` stays the only explicit free point for trace-owned refs; the shared-`seen` dedup across subtrees is kept (now covered by `test_trie_release_frees_each_ref_once_across_shared_subtrees`), and `test_trie_overwrite_drops_handle_without_freeing_borrowed_refs` asserts overwrite calls no free while `ray.get` on the borrower's handle still succeeds.

xtuner/v1/rl/rollout/utils.py
```diff
@@ -209,6 +216,19 @@ async def postprocess(
                 f"prompt_tokens={prompt_tokens}, completion_tokens={completion_tokens}"
             )
             rollout_state.routed_experts = ray.put(concat_routed_experts)
+            rollout_state.routed_experts_owner = "rollout"
+            if release_input_routed_experts:
+                if history_routed_experts_owner == "rollout":
+                    free_object_refs(
+                        history_routed_experts_ref
+                        if isinstance(history_routed_experts_ref, list)
+                        else [history_routed_experts_ref]
```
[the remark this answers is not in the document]
matrix72c: Fixed. The inconsistency was resolved as a side effect of folding release into `reset`/`discard` in 0ad10e5: `postprocess` now frees the history ref whenever `routed_experts_owner != "trace_store"`, i.e. the same "anything not explicitly borrowed is released" rule as `rl_data.py`, so an untagged (legacy-checkpoint restore) history ref is released after concatenation instead of being overwritten and leaked. Added `test_postprocess_frees_unowned_history_ref` in a4e94b16 covering the `owner=None` branch.

YanhuiDua (approved):
LGTM, But going forward, we need to abstract routed experts as a separate distributed `RoutedExpertsStore`. @YifanHe-ailab

xtuner/v1/rl/rollout/utils.py
matrix72c: 已删除

PR #2066, opened 2026-09-04
[ci] add multi nodes case for rl

kkscilife:
Don't support tp_size.
 Don't support sync_weights_interval>1 on colocated mode


zhulinJulia24 (approved):
LGTM

PR #2069, opened 2026-09-04
fix(muon): avoid EP/FSDP AGRS OOM

jayhenry:
## Summary

Fixes an EP+FSDP AGRS out-of-memory in the Muon optimizer's remainder-batch path, then replaces the
implicit remainder handling with an explicit, benchmarked `remainder_strategy` choice
(`agrs` / `pad_all2all` / `ragged_all_to_all`) and a companion sub-group all-gather fix.

## Root Cause

For GLM-5.2 with 512 ranks and EP8, the expert FSDP group has 64 ranks. The 76 same-shaped
`fused_w2` parameters form a remainder batch of 12. AGRS used the full DTensor dimension
`(256 * 6144)` instead of the EP-local, FSDP-visible dimension `(256 * 6144) / 8`, producing a
72 GiB communication buffer and a second 72 GiB `flatten()` copy.

AGRS itself also all-gathers every matrix in a batch onto every rank, so its time and peak memory
scale linearly with batch size — worst case is exactly a large remainder batch, which is where
GLM-5.2 hit this OOM.

## Fix

- **FSDP-visible dimension**: compute the EP-adjusted global shard dimension once per
  shape-homogeneous batch instead of once per mesh group (the previous scope let a later
  differently-shaped batch reuse an earlier batch's dimension). `Muon._create_muon_tasks` was
  split into `_resolve_mesh_comm_plan` (mesh-topology / communication-strategy, resolved once per
  `(device_mesh, placements)` group) and `_build_muon_task` (per-batch, shape-dependent metadata),
  so the split makes this scoping explicit instead of implicit in one long function.
- **Sub-group all-gather guard**: the sub-group path (each MoE expert spans a small rank
  sub-group) was gated on `len(params) < fsdp_size`, so a model whose expert params happen to
  fill a batch (GLM-5.2: 76 fused expert weights over 64 ranks) fell back to full-group
  all-to-all / AGRS reconstructing the whole multi-expert matrix on every rank. The guard is
  dropped so any group whose experts span a sub-group always takes the cheaper path.
- **`remainder_strategy` option**, benchmarked at world size 8 on 96 MB (8192x6144, bf16)
  matrices:
  - `"agrs"` — all-gather + reduce-scatter; time/memory grow linearly with remainder size
    (14.9 ms / 624 MB at R=1 → 20.1 ms / 2016 MB at R=7).
  - `"pad_all2all"` (default) — zero-pads the remainder to world size and uses a uniform
    all-to-all; flat ~15.2 ms regardless of R, but every rank (including idle ones) allocates a
    full matrix.
  - `"ragged_all_to_all"` — sends only the real matrices via all-to-all split sizes, so idle
    ranks allocate nothing (648 MB busy / 48 MB idle at R=2, vs. 804 MB / 804 MB for padded
    all-to-all). The same split-size path also covers uneven shards from DTensor's
    `ceil(size / world_size)` chunking, so it needs no separate padding branch.
  - `enable_all2all=False` now coerces the strategy to `"agrs"` instead of rejecting the
    combination, since AGRS is the only remainder path that doesn't use all-to-all.
- Removed the scratch OOM-investigation notes and repro script now that the fix has a permanent
  regression test.

## Test Plan

- `PYTHONPATH=. python -m pytest tests/optim/test_muon.py -q` — 20 passed (run under the local
  GPU lock), covering:
  - FSDP parity for all three remainder strategies against a single-process reference.
  - EP+FSDP batch-specific FSDP-visible-dimension regression (fails on the pre-fix revision,
    where Newton-Schulz sees the un-divided global shape instead of the EP-local one).
  - Sub-group all-gather is taken even when expert params fill a full batch.
  - `remainder_strategy` validation and default-value behavior, including the
    `enable_all2all=False` coercion.
- `ruff check` / `ruff format --check` on the changed files.


xtuner/v1/config/optim.py
```diff
@@ -78,6 +78,10 @@ class MuonConfig(OptimConfig):
     enable_all2all: Annotated[
         bool, Parameter(help="Allow all-to-all comm strategy; set False on topologies where all-to-all is unreliable")
     ] = True
+    remainder_strategy: Annotated[
+        Literal["agrs", "pad_all2all"],
+        Parameter(help="Communication strategy for Muon parameter batches smaller than the FSDP group size"),
+    ] = "agrs"
```
nil0x9, line 84: LGTM. It's good to have another option for flexibility.

xtuner/v1/optim/muon.py
```diff
@@ -716,6 +722,12 @@ def _create_muon_tasks(
                     batch_size=group_world_size,
                     extra_group_key=self._muon_split_sizes.get,
                 ):
+                    global_shard_dim_size: int | None = None
+                    if sharded_tensor_dim is not None and non_fsdp_shard_factor > 1:
+                        full_shard_dim_size = params[0].size(sharded_tensor_dim)
+                        assert full_shard_dim_size % non_fsdp_shard_factor == 0
+                        global_shard_dim_size = full_shard_dim_size // non_fsdp_shard_factor
```
nil0x9, line 729: nice catch

xtuner_muon_oom.md
```diff
@@ -0,0 +1,210 @@
+# GLM-5.2 512 卡 Muon 更新阶段 OOM
+
+## 结论
+
+这次 OOM 的根因是 **EP 与 FSDP 同时沿参数第 0 维分片时，AGRS 使用了错误的“全局维度”**。`DTensor.size(0)` 是包含 EP 的完整模型维度，但 AGRS 的 FSDP process group 只能在固定 EP rank 内恢复参数，因此它应该使用 **EP-local global dimension**。修复前代码直接把前者传给 AGRS，令 padding 行数放大为正确值的 8 倍；本分支已加入临时绕过和永久修复。
```
nil0x9, line 5: 抱歉又留了个坑，之前只在 qwen3.5 的有限 setting 下测，没暴露问题。修正 global dim 确实可以解决这个问题，但是我觉得根因不在这。在之前的设计中，MoE 部分的参数是尽量要避免通信的，按照 XTuner 的参数切分逻辑， `num_experts % world_size == 0` 的时候，刚好每张卡持有完整的 1-n 个专家权重，0 通信，`world_size % num_experts == 0` 的时候，几张卡共同持有一个专家，这时候需要在一个比较小的通信组中做 allgather。原则上大部分情况下不应该触发 AGRS 这条通路-- 这个是给 non-MoE 参数的，开启 ep 的时候在 ep mesh 上是 Replicate，不会触发这个 PR 修复的 global dim 的 bug。

如果我理解得没错， 真正问题是这里
https://github.com/InternLM/xtuner/blob/7d37742425f53205e61757f0853074b0606664ea/xtuner/v1/optim/muon.py#L684-L690

这里 `and len(mesh_params) < fsdp_size` 的条件太苛刻了，在 GLM 的层数下问题爆发了（导致 MoE 走了原来 non-MoE 参数应该走的通信路径），实际上 MoE 无论层数多少都可以走小通信组的 allgather，可以分 batch 一部分走 all2all，一部分走小通信组 allgather。把这个条件去掉，做一些更细粒度的 batch 调度可能是更优的解法。

sallyjunjun (approved):

PR #2071, opened 2026-09-05
[Fix] Update flush cache worker broadcast

matrix72c:
## Summary
- update `RolloutController.flush_cache()` to use the lifecycle-aware `_broadcast_to_workers` helper after the worker broadcast refactor
- add a regression test covering the active-worker flush-cache dispatch

## Root cause
The upstream flush-cache feature added a call to `_broadcast_to_active_workers`, while the later controller refactor renamed that helper to `_broadcast_to_workers` and missed this caller.

## Test plan
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=.../xtuner sandbox_rl/.venv/bin/python -m pytest -q tests/rl/test_rollout_logic.py` (61 passed)

### DIFF
--- a/tests/rl/test_rollout_logic.py
+++ b/tests/rl/test_rollout_logic.py
@@ -455,6 +455,14 @@ class TestRolloutController(unittest.IsolatedAsyncioTestCase):
         with self.assertRaisesRegex(TimeoutError, "rollout worker recovery"):
             controller.pause_generation()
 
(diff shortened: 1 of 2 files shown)


YanhuiDua:
最新 main 分支已经删掉了 flush_cache，我理解这个 PR 的修改就不需要了

matrix72c:
ok，这个pr我关闭

PR #2088, opened 2026-09-15
fix(activation-offload): clear runtime state after train step

Denny991:
## What does this PR do?

Fixes a constant cross-step GPU memory floor in GLM-5.2 training when DSA Top-K activation offload is enabled.

## Why is this needed?

The saved Top-K tensors are stored in the process-global `OffloadManager`. After the backward pass of the final offload window, the runtime entries (`items`, `may_npu_tensors`, `getcnt`) were not cleared in the SFT training path. The highest-block key stays in `items` (no higher layer prefetches it), and block-0 keys are moved into `may_npu_tensors` by prefetch but never deleted. These entries hold strong references to `SwapTensor.tensor` (GPU, restored to full size by `launch_h2d`) and `tensor_cpu` (pinned), keeping them alive until the next step's forward overwrites them via `put()`.

This creates a **constant ~128 MiB floor** (GLM-5.2 16K packed) that persists across the step boundary, overlapping the `clip_grad_norm` + `step_optimizer` memory peak — the worst possible position. `gc.collect()` does not help because these are strong references, not reference cycles.

## What has changed?

- Add a step-scoped `OffloadManager.clear_step()` lifecycle API.
- Use `torch.cuda.synchronize()` to drain all pending offload copies before releasing runtime entries — offload streams are owned by external subsystems (FSDP, MoE) and the manager does not track them.
- Clear offload runtime tensors and counters after all microbatch backwards complete, while retaining reusable pinned CPU buffers.
- Wrap the `train_step` microbatch loop in `try/finally` so cleanup runs even when backward raises (OOM retry, NaN abort).
- Remove the now-redundant `OffloadManager().clear()` call from the RL training worker.
- Add unit tests covering runtime cleanup, group isolation, pinned-buffer retention, and no-op behavior without offload.
- Add real GPU-path regression assertions in `test_glm52_mtp_checkpoint_repro` verifying that runtime entries are released after `train_step` while pinned buffers stay cached.

## Validation

- `pytest tests/utils/test_activation_offload.py -q`: 2 passed
- `pytest tests/model/test_glm52_mtp_checkpoint_repro.py -v`: 3 passed (GPU)
- `pytest tests/utils/ -v --ignore=tests/utils/test_internal_metrics.py`: 75 passed
- Ruff check / format / docformatter / py_compile: all passed
- GLM-5.2 8-GPU H200 run with `recompute_ratio=0.25`, DSA Top-K offload, DeepEP, and CUTLASS for 35 steps (full epoch): all 8 ranks completed successfully, no OOM.
- Rank0 `max_memory` stayed flat at ~101-102 GiB from step 2 through step 35, with zero growth trend. (Before this fix, the same configuration showed monotonically increasing memory across steps and OOM when running longer.)
- Step time comparable to the per-stream sync approach (~4100 tgs average), confirming `torch.cuda.synchronize()` at the step boundary is not a performance regression.

## Notes

The fix is applied after the complete `TrainEngine.train_step()` backward path, before gradient clipping and optimizer update. No checkpoint or optimizer state behavior is changed.


jayhenry:
### Suggestion: the stream registry (`register_stream` / `synchronize` / `offload_streams`) can be dropped entirely

The fix direction is right, and the `TrainEngine` contact point is nicely minimal (one named call, no new field, no branching). One design point though: the three new members exist only to answer a single question — *which streams must drain before we drop the `SwapTensor` references?*

`TrainEngine` is the one that triggers cleanup, but it does not know which stream offload uses (it is owned by `MoE.offload_stream` / FSDP's `all_gather_stream` and reaches the hook through `async_save_on_cpu.__init__`). The registry's entire job is to carry that stream from the construction site to the cleanup site.

**To be clear, the synchronization itself is not redundant defensive code — the dangling risk is real:**

- `SwapTensor.launch_d2h()` (`activation_offload.py:86-99`) reads `self.tensor` on the d2h stream with **no** `record_stream(d2h_stream)`. Freeing that GPU block while the copy is in flight lets the caching allocator hand it to the compute stream, so the D2H can read corrupted data.
- `SwapTensor.prefetch_launch_h2d()` (`activation_offload.py:141-159`) reads `self.tensor_cpu` (pinned) on the h2d stream. The GPU side is covered by `record_stream`, but **the pinned CPU buffer is not**. At the call sites with `reserve_pin_memory=False` (`xtuner/v1/model/moe/qwen3vl_text.py:144` and both vision encoders), `tensor_cpu` is owned solely by the `SwapTensor`, so dropping the reference frees pinned memory under an in-flight copy.

**So the thing to remove is the registry, not the sync.** At a step boundary, "wait on these two streams" and "wait on the device" cost effectively the same: the next statement is `total_loss.item()`, which already blocks the host, followed by the grad-norm all-reduce and the optimizer step. Stream-precise waiting buys nothing here.

```python
# activation_offload.py — drop offload_streams / register_stream / synchronize / has_runtime_state
def clear_step(self, group: str | None = None) -> None:
    """Release saved-tensor offload state at a training-step boundary.

    Pinned CPU buffers stay cached for reuse; GPU tensors and per-step counters are released.

    Args:
        group (str | None): Optional group to clear. If omitted, clear all groups.
    """
    if not (self.items or self.may_npu_tensors):
        return
    # Offload copies run on streams this manager does not own. Drain the device before
    # dropping references: launch_d2h reads tensor on the d2h stream without record_stream,
    # and prefetch_launch_h2d reads the pinned CPU buffer, which record_stream cannot cover.
    torch.cuda.synchronize()
    self.clear(group=group)
```

```diff
 # async_save_on_cpu.__init__
-        OffloadManager().register_stream(group, h2d_stream)
-        OffloadManager().register_stream(group, d2h_stream)
```

No NPU special-casing is needed: `transfer_to_npu` at the top of the file already redirects `torch.cuda.*`, consistent with the existing `torch.cuda.Event` / `Stream` usage in this module.

**What this also fixes for free:**

| Issue in the current diff | After dropping the registry |
| --- | --- |
| `tests/utils/test_activation_offload.py` leaves a `mock.Mock` permanently in the process-wide singleton's `offload_streams` — the `finally` block's `clear(clear_pin_memory_cache=True)` does not clear that field, so later tests in the same pytest session (including the GPU cases that now reach `clear_step()`) will call into the mock | Gone. With no streams to register, `clear()` is a complete rollback and the test becomes a pure state assertion |
| `offload_streams` is append-only: none of `clear()`'s three paths touch it, so a rebuilt engine (RL worker reloading a model) keeps stale streams alive and synchronizes them every step | Gone |
| The manager registers and hard-synchronizes FSDP's internal `all_gather_stream` (`qwen3vl_text.py:143`, default group) — a stream whose lifetime, reshard-after-backward behaviour and `unshard` reuse are not `OffloadManager`'s to reason about | Gone |
| `has_runtime_state()` currently guards nothing: when offload is off, `offload_streams` is an empty dict, so `synchronize()` is a zero-cost empty loop and `clear()` just clears empty dicts | **Inverted** — with a device sync the guard now prevents a real full-device stall, so it is worth keeping, inlined as `if not (self.items or self.may_npu_tensors)` |
| `stream: Any` / `dict[str, list[Any]]` weaken strict mypy for `xtuner/v1/*` (the caller already types these as `torch.cuda.Stream`) | Gone along with the field |

Net: +1 public method instead of +4, no new field, and the `TrainEngine` contact point is unchanged (still one `clear_step()` call).

**The one cost:** `torch.cuda.synchronize()` is a stronger barrier than waiting on two copy streams — it also waits on FSDP's communication streams. Right after backward and before gradient clipping this should be within noise, but a 60-step step-time comparison would settle it. If it does regress, the fallback is to record a D2H-completion event on `SwapTensor` and have `clear_step()` wait on those events — that approach does not need a registry either.

---

Two smaller points while you are in here:

1. **`clear_step()` is not in a `try/finally`.** If backward raises (OOM retry, NaN abort), cleanup is skipped and the stale state carries into the next step. Either wrap the microbatch loop in `try/finally`, or expose the lifecycle as a scope (`@contextlib.contextmanager def step_scope(...)`) so the boundary is self-documenting. Orthogonal to the suggestion above.
2. **`xtuner/v1/rl/trainer/worker.py:908`** still calls `OffloadManager().clear()` right after `train_step()`. That is now redundant — and it is the less correct version, since it does not synchronize first. Worth removing (along with the import at line 60) so there is a single owner of step-boundary cleanup.

jayhenry (approved):

PR #2095, opened 2026-09-17
[Fix] Align RL sample advantage length and position with packed input_ids

YanhuiDua:
## Root Cause

In `BaseRLTrainer._prepare_train_data` (token-in-token-out branch), the per-sample advantage list was built as `[adv] * len(prompt_ids) + [mask-based adv]`, which is **one element longer** than `input_ids` (`prompt_ids + response_ids[:-1]`). Since `advantage[i]` must align with `shifted_labels[i]`, every sample contributed one extra element. After `TrainingController._packing` concatenated multiple samples into one sequence, the packed `advantages` tensor was longer than `shifted_labels` and position-shifted, so the element-wise policy loss (`-ratio * advantages`) either crashed on shape mismatch or - with sequence-parallel padding - silently misaligned.

The existing tests asserted this buggy layout (e.g. `len(advantage) == shifted_labels.numel() + 1`), which is also corrected here.

## Fix

- `xtuner/v1/train/rl_trainer.py`: prompt-position advantages are now `0.0` (their shifted labels are -100 and masked out of the loss), and the list length equals `len(input_ids)`; the metrics-side compensating `[:-1]` is removed accordingly.
- `xtuner/v1/train/rl_trainer.py`: `advantages/mean|min|max` statistics now count only loss-active tokens (`label != -100` in the agentic branch, `mask != 0` in the token-in-token-out branch). Previously the full layout - including prompt-position `0.0` placeholders and masked-out response tokens - diluted the statistics.
- `xtuner/v1/rl/trainer/controller.py`: added `_verify_packed_alignment`, which asserts after packing that `shifted_labels`, `advantages` and, when present, `rollout_logprobs` / `teacher_logprobs` / `target_token_ids` / `teacher_indices` all match the `input_ids` length - mismatches now fail fast at pack time instead of entering the policy loss. Added the `PackedBatch` TypedDict describing the packed structure. The asserts are shape-only checks, negligible overhead on the packing path.

## Test Plan

- `tests/rl/test_prepare_train_data.py`: fixed three existing tests that encoded the old buggy advantage layout.
- Added `TestPrepareTrainDataPackAlignment` (no-padding / with-padding / teacher-fields cases): packs 4 samples from 2 groups with deliberately varied prompt/response lengths, masks and advantage values, then verifies `input_ids`, `shifted_labels`, `advantages`, `rollout_logprobs` (and teacher fields) position-by-position against the expected per-sample layout, plus the pad-tail contract (labels -100, advantages -100).
- Added `test_advantage_stats_count_only_loss_active_tokens`: one group mixing a token-in-token-out sample (advantage 2.0, mask `[1,0,1,0]` -> 2 active tokens) with an agentic sample (advantage -1.0, 2 non `-100` labels); asserts `advantages/mean == 0.5`, `min == -1.0`, `max == 2.0`. Verified it fails under the old statistics scope.
- Verified the new tests fail on the unfixed code (reproducing the bug) and pass with the fix; `pre-commit run` (ruff, ruff-format, mypy strict, docformatter) passes.

ProduceBatchResult impact: not affected (batch accounting, timing and reward metrics are unchanged; `advantages/mean|min|max` are trainer info-dict statistics and only their token scope narrows to loss-active tokens).

RoutedExperts impact: not affected.

Ray concurrency impact: not affected.


xtuner/v1/train/rl_trainer.py
PengchengShi00, line 1394: 这里是对所有的token的advantages去做统计，需要mask一下么，如果这个统计指标不是很重要的话，也无所谓

YanhuiDua: 新的写法改成了prompt部分adv_val是0
` actual_advantages = [0.0] * (len(prompt_ids) - 1) + [ 0.0 if mask == 0 else advatnages_val for mask in response_mask]`，所以这个指标是准确的

YanhuiDua: done

PengchengShi00 (approved):
LGTM

